PR #4, opened 2020-12-02
Add localization

sprucely:
i18next requires a bit of ceremony to get configured, but it seems to be the most popular choice.


jSylvestre (approved):
Didn't test it, but looks good to me.

HarvestMobile/App.tsx
```diff
@@ -1,19 +1,31 @@
...
 export default function App() {
   const isLoadingComplete = useCachedResources();
   const colorScheme = useColorScheme();
+  const isI18nInitialized = useState(false);
 
-  if (!isLoadingComplete) {
+  useEffect(() => {
+    i18n.init()
+      .then(() => {
+        isI18nInitialized.set(true);
```
srkirkland, line 23: is this a thing now instead of doing `const [isI18nInitialized, setIsI18nInitialized] = useState(false)`?

srkirkland (approved):

HarvestMobile/App.tsx
sprucely: It is for hook-state. Unfortunate that they didn't choose a different name, such as `useHookState`.

srkirkland: oh that sucks

PR #6, opened 2020-12-11
Add Serilog console logging

sprucely:
All the examples for using scoped logging involved `Dictionary<string, object>` literals passed into `logger.BeginScope()`. I find dictionary literals quite ugly, so I added some extension methods to make the api similar to Serilog's `logger.ForContext()`


Harvest.Web/Helpers/LogConfiguration.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Harvest.Web.Helpers
+{
+    public static class LogConfiguration
+    {
+        private static bool _loggingSetup;
+
+        private static IConfiguration _configuration;
+
+        public static void Setup(IConfiguration configuration)
```
srkirkland, line 19: @sprucely is all of this recommended or is this just copied from an old project?  On my newer projects I've left most of this out.  is it doing something important.

srkirkland (commented):
While I think the extensions are clever I like it less than the old Log.Information() style logging in just about every way. I guess maybe there is some benefit to DI for the ILogger but I can't think of it.

Harvest.Web/Startup.cs
```diff
@@ -55,8 +57,11 @@ public void ConfigureServices(IServiceCollection services)
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerFactory loggerFactory)
         {
+            LogConfiguration.Setup(Configuration);
+            loggerFactory.AddSerilog();
```
srkirkland, line 63: I didn't think this would still be a thing anymore -- I didn't recall seeing it in the more modern examples. what does it do?

Harvest.Web/Helpers/LogConfiguration.cs
sprucely: Just about all examples I've seen use appsettings.json for Serilog config. But I copied from Peaks, since I thought it was the current/preferred way for our systems. I can go back and change it.

Harvest.Web/Startup.cs
sprucely: Copied from Peaks. Apparently nothing if you don't provide a customized Serilog ILogger argument. I'll remove it.

sprucely:
> 
> 
> While I think the extensions are clever I like it less than the old Log.Information() style logging in just about every way. I guess maybe there is some benefit to DI for the ILogger but I can't think of it.

I don't follow. How is this any worse than the equivalent in Peaks...
```
var logger = Log.Logger
    .ForContext("source", "browser")
    .ForContext("messageBody", message, true);
logger.Information(message.Message);
```
The only thing I can think of is the disposable blocks adding a bit of noise, but that was only necessary for the example's back-to-back scopes. I thought it would be good to avoid `Log.Logger` because of the pain I experienced when AutoMapper dropped their global static config system.

srkirkland:
> > While I think the extensions are clever I like it less than the old Log.Information() style logging in just about every way. I guess maybe there is some benefit to DI for the ILogger but I can't think of it.
> 
> I don't follow. How is this any worse than the equivalent in Peaks...
> 
> ```
> var logger = Log.Logger
>     .ForContext("source", "browser")
>     .ForContext("messageBody", message, true);
> logger.Information(message.Message);
> ```
> 
> The only thing I can think of is the disposable blocks adding a bit of noise, but that was only necessary for the example's back-to-back scopes. I thought it would be good to avoid `Log.Logger` because of the pain I experienced when AutoMapper dropped their global static config system.

Yea, that's exactly my thought -- I think the disposable blocks add noise and the need to inject in the logger adds noise.  Not saying it won't be worthwhile noise but I was hoping to see cleaner syntax and I just didn't see anything that I liked more that what we've been using for all these years.

Harvest.Web/Helpers/LogConfiguration.cs
srkirkland: oh I don't mean the appsettings stuff when I said is all of this recommended, I mean the stuff in this method, like
```            
AppDomain.CurrentDomain.UnhandledException += (sender, e) => Log.Fatal(e.ExceptionObject as Exception, e.ExceptionObject.ToString());

AppDomain.CurrentDomain.ProcessExit += (sender, e) => Log.CloseAndFlush();

#if DEBUG
            Serilog.Debugging.SelfLog.Enable(msg => Debug.WriteLine(msg));
#endif
            _loggingSetup = true;
```

sprucely:
Ugh, I thought the new using statement syntax would be cleaner. It is, a bit. But it requires assigning the disposable to a variable...
```
using var _ = _logger
    .WithScope("SomeScope", 1)
    .WithScope("AnotherScope", "Hello")
    .BeginScope();
```

Is that acceptable, or should I go ahead and remove those extensions? I'm thinking remove, because there are other Serilog enrichment patterns not possible when using  AspNetCore's ILogger.

srkirkland:
I'll have to think about it a little bit -- the syntax is not bad at all, and there is some benefit to going vendor neutral instead of using Serilog Log.* directives.  I'd say probably remove it for now and save the code in case we want it later, or just wait and when jason gets back we can talk about it as a group

Harvest.Web/Middleware/SerilogHelpers.cs
```diff
@@ -0,0 +1,47 @@
...
+                    : LogEventLevel.Information;
+
+        public static void EnrichFromRequest(IDiagnosticContext diagnosticContext, HttpContext httpContext)
+        {
+            var request = httpContext.Request;
+
+            // Set all the common properties available for every request
+            diagnosticContext.Set("Host", request.Host);
+            diagnosticContext.Set("Protocol", request.Protocol);
+            diagnosticContext.Set("Scheme", request.Scheme);
```
srkirkland, line 28: Why sometimes set the context info here and sometimes via the middleware IActionFilter?  Seems to me the action filter could do a fine job setting host/protocol/scheme and other vars just as well as route and action info?

Harvest.Web/Program.cs
```diff
@@ -1,23 +1,52 @@
...
+#if DEBUG
+            Serilog.Debugging.SelfLog.Enable(msg => Debug.WriteLine(msg));
+#endif
+            // provide some basic console logging (can be captured by Azure logging)
+            Log.Logger = new LoggerConfiguration()
+                .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
+                .MinimumLevel.Override("Microsoft.AspNetCore", LogEventLevel.Warning)
+                .Enrich.FromLogContext()
+                .WriteTo.Console()
+                .CreateLogger();
```
srkirkland, line 28: I do like this idea of basically creating a throwaway logger that just writes to console just to catch startup errors.

Harvest.Web/appsettings.json
```diff
@@ -1,10 +1,28 @@
 {
-  "Logging": {
-    "LogLevel": {
+  "Serilog": {
```
srkirkland, line 2: I'm getting flashbacks of Nlog/Log4Net here it's burning my eyeballs 🔥

sprucely: What, you prefer strongly-typed code over json config?

srkirkland: ha yup this isn't Java Spring! 😁

Harvest.Web/Middleware/SerilogHelpers.cs
sprucely: Doh! I didn't even notice that both HttpContext and DiagnosticContext are available in EnrichFromRequest.

sprucely: Never mind. I think it has to stay this way.  I don't see any way to get at ActionExecutingContext from the HttpContext.

srkirkland: Yea makes sense, sounds like it would be better to move this into an action filter since you can't do it the other way around.  And maybe if we get an action filter we really like that adds all of our regular serilog context stuff we can make a nuget package out of it.  I would love to get this sort of boilerplate out of a single project and into a common lib.  Stuff like all of this info, user info, and client IP/browser info would be great to have either as an enricher, action filter, or just some middleware.

srkirkland (approved):
very nice i like the way this turned out

PR #7, opened 2020-12-17
Add AppDbContext, domain types and migrations

sprucely:


Harvest.Core/Domain/Project.cs
```diff
@@ -0,0 +1,110 @@
...
+            modelBuilder.Entity<Project>().HasIndex(a => a.QuoteId);
+
+            modelBuilder.Entity<Project>().Property(a => a.ChargedTotal).HasPrecision(18, 2);
+            modelBuilder.Entity<Project>().Property(a => a.QuoteTotal).HasPrecision(18, 2);
+
+            modelBuilder.Entity<Account>()
+                .HasOne(a => a.Project)
+                .WithMany(p => p.Accounts)
+                .HasForeignKey(a => a.ProjectId)
+                .OnDelete(DeleteBehavior.Restrict);
```
srkirkland, line 88: I don't know that there are required but I actually really like them -- makes the relationships explicit.

srkirkland (approved):
nice!

PR #8, opened 2021-01-05
Configure jest and add client testing to azure pipeline

sprucely:
This introduces a dependency on rescripts which enables overriding CRA's internal configs without ejecting. I used it to configure `jest-trx-results-processor`.

I'm setting coverage thresholds at 0 for now just to see how useful the info is after we get some actual code in there.

Also updated solution so that visual studio displays some of the files in the root folder.


Harvest.Web/ClientApp/package.json
```diff
@@ -16,11 +16,16 @@
...
+    "@rescripts/cli": "0.0.15",
+    "cross-env": "^7.0.3",
+    "jest-trx-results-processor": "^2.2.0"
+  },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "start": "rescripts  start",
```
jSylvestre, line 25: Is rescripts just informational? I'd say react-scripts is more descriptive

srkirkland: yea i agree I don't like that we had to add a whole rescripts cli package just to get code coverage to work.  Reading the repo for rescripts it sounds like something I'd like to avoid having

srkirkland: I see why you had to do it since CRA doesn't support the results processor but hopefully there is another option

srkirkland: Can you just add the reporter in the CLI `npm run citest` config directly, like `--reports jest-trx-results-processor` or something?

sprucely: I had done it that way because there doesn't appear to be a way to override default configs for `jest-trx-results-processor` via cli. I wanted both trx and coverage files to be located in the output folder. They're now being placed in the ClientApp folder.

srkirkland:
@sprucely we should just both merge to main

PR #9, opened 2021-01-29
add elastic logging

srkirkland:


Harvest.Web/Program.cs
```diff
@@ -19,13 +21,44 @@ public static int Main(string[] args)
...
                 .MinimumLevel.Override("Microsoft.AspNetCore", LogEventLevel.Warning)
                 .Enrich.FromLogContext()
-                .WriteTo.Console()
-                .CreateLogger();
+                .Enrich.WithClientIp()
+                .Enrich.WithClientAgent()
+                .Enrich.WithExceptionDetails()
+                .Enrich.WithProperty("Application", loggingSection.GetValue<string>("AppName"))
+                .Enrich.WithProperty("AppEnvironment", loggingSection.GetValue<string>("Environment"))
+                .WriteTo.Console();
```
sprucely, line 49: Didn't notice this before, but is there a reason we have a `LoggerConfiguration` initialized both here and in SerilogHelpers.cs?

srkirkland: @sprucely it's a good question and my guess is no, we can probably dump almost all traces of it from the Startup.cs because we want to be initializing the logs as early as possible and thus Program.cs is the way to go.  But I didn't want to mess with what you did in startup so I left it.  Maybe I'll setup a db and give it a full test with and without before merging

sprucely (approved):
Works for me. Middleware is logging to elastic, though no user details, since it is only static content so far.

sprucely (approved):
Looks good.

PR #15, opened 2021-03-12
Use db Initialize to create any missing roles on startup

jSylvestre:


sprucely (approved):

Harvest.Core/Data/DbInitializer.cs
```diff
@@ -15,9 +16,32 @@ public DbInitializer(AppDbContext dbContext)
...
+        private async Task CheckCreateRole(string role)
+        {
+            if (!await _dbContext.Roles.AnyAsync(a => a.Name == role))
+            {
+                var roleToCreate = new Role {Name = role};
+                await _dbContext.Roles.AddAsync(roleToCreate);
+                await _dbContext.SaveChangesAsync();
+            }
+            
+            return;
```
srkirkland, line 44: @jSylvestre suggestion: do the SaveChangesAsync() outside of this method.  No reason to do multiple saves when you can batch saves together, and it'll only get more important as the init method grows

srkirkland (approved):

PR #21, opened 2021-03-12
Jcs/iam service

jSylvestre:


Harvest.Web/Services/UserService.cs
```diff
@@ -27,22 +27,22 @@ public UserService(AppDbContext dbContext, IHttpContextAccessor httpContextAcces
         public async Task<User> GetCurrentUser()
         {
             var username = _httpContextAccessor.HttpContext.User.Identity.Name;
+            var userClaims = _httpContextAccessor.HttpContext.User.Claims.ToArray();
+            string iamId = userClaims.Single(c => c.Type == "ucdPersonIAMID").Value;
 
-            var dbUser = await _dbContext.Users.SingleOrDefaultAsync();
+            var dbUser = await _dbContext.Users.SingleOrDefaultAsync(a => a.Kerberos == username && a.Iam == iamId);
```
srkirkland, line 33: @jSylvestre  If usernames can change i guess we should just not check by kerb at all?

jSylvestre: @srkirkland Now that we can get IAM ids from CAS, that would be my preference. (But, it is still rare, happened twice for OPP that I know of)
But I do like including the kerb in the database. Easier to read

srkirkland: yea i think it's useful in the db to read and also for debugging.  but in the app for any comparison to db query we should probably leave it out

PR #22, opened 2021-03-12
take eduPerson scope and use it to populate IAMID

srkirkland:


sprucely (approved):
So CAS does provide IamId after all.

srkirkland:
> So CAS does provide IamId after all.

Yes but only via OAuth.  And we are fortunate too because this was not in it originally but Travis added it for me which will turn out to be very helpful.

jSylvestre (approved):

jSylvestre:
Once this merges, I'll pull it into my PR and undo my changes

PR #27, opened 2021-03-12
User Permissions

jSylvestre:
Ugly, but works:

image

### DIFF
--- a/Harvest.Web/Controllers/PermissionsController.cs
+++ b/Harvest.Web/Controllers/PermissionsController.cs
@@ -26,7 +26,7 @@ namespace Harvest.Web.Controllers
         }
         public async Task<IActionResult> Index()
         {
-            //TODO: Filter out System roles?
+            //TODO: Filter out Role.Codes.System roles?
             var permissions = await _dbContext.Permissions
                 .Include(a => a.User)
                 .Include(a => a.Role)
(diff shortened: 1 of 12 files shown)


Harvest.Web/Controllers/PermissionsController.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> AddUserRole(AddUserRolesModel model)
+        {
+            var viewModel = new AddUserRolesModel
+            {
+                Roles = await _dbContext.Roles.Where(a => a.Name != "System").ToListAsync(),
```
srkirkland, line 44: Remember to update this once Spruce merges the non magic string version

srkirkland (approved):

Harvest.Web/Controllers/PermissionsController.cs
jSylvestre: @srkirkland I was reserving System for us. As this page allows Admin access, I don't want system as a choice

srkirkland: @jSylvestre Yes but I'm saying the string "System" you should change to RoleCodes.System so it's not just a string

Issue #17, opened 2021-03-12
Add maps to react


srkirkland:
Likely using mapbox and leaflet.  Determine what libraries would be needed to allow "drawing" and labeling plots on the map.

PR #28, opened 2021-03-12
Throw in a map with the ability to draw regions

srkirkland:
Testing out adding in a map and drawing polygons on it.  Worked pretty well.  Uses leaflet and OSM.  I don't know if we want to merge because we might end up using something different for the drawing part, but for the map itself i'm pretty sure it'll be leaflet for the interaction layer.

The tiles themselves can be swapped out, if we aren't happy with these free basic ones Cal can design us some cool map layers.

image

### DIFF
--- a/Harvest.Web/ClientApp/package-lock.json
+++ b/Harvest.Web/ClientApp/package-lock.json
@@ -1,8 +1,19144 @@
 {
   "name": "harvest.web",
   "version": "0.1.0",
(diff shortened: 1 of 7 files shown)


srkirkland:
let's say this closes #17

sprucely (approved):
Worked for me.
image

jSylvestre (approved):

PR #31, opened 2021-03-12
Add SuperController

### LINKED ISSUE #24: Create a SuperController
Primarily for storing message/errorMessage in TempData

sprucely:
closes #24

### DIFF
--- a/Harvest.Web/Controllers/HomeController.cs
+++ b/Harvest.Web/Controllers/HomeController.cs
@@ -6,7 +6,7 @@ using Microsoft.AspNetCore.Mvc;
 namespace Harvest.Web.Controllers
 {
     [Authorize]
-    public class HomeController : Controller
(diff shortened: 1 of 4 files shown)


srkirkland (approved):
You'll eventually probably want some sort of antiforgery token attribute but i'm actually happy you left it off.  We can decide if we actually do want it on all controllers or maybe just some subset.

jSylvestre (changes requested):
I think you need to change the layout page as well to display this info.

sprucely:
> 
> 
> I think you need to change the layout page as well to display this info.

Already present...
https://github.com/ucdavis/Harvest/blob/6d4d327f9a1c085ff555db5c0f3c64840304f3c7/Harvest.Web/Views/Shared/_Layout.cshtml#L69-L88

PR #38, opened 2021-03-15
Add jquery and datatables

### LINKED ISSUE #18: Import the latest datatables
Similar to peaks and most of our other apps, we need to import datatables, including the styles for datatables to look good with bootstrap.  We should import those from the correct CDNs and be able to only load them on pages that need them.

sprucely:
I included the datatables responsive plugin, which I believe will become useful as more columns get added to reports.

closes #18


srkirkland (approved):
i'm not sure we need jquery to load on every page but it's also not a problem at this stage and we can change it later if need be.  I'd say 🚢

PR #44, opened 2021-03-18
Remove Permissions

jSylvestre:


sprucely:
Is there something in place to prevent a department admin from assigning system admin?

jSylvestre:
@sprucely Yes, I filter out the System role. Could also add a check there it is actually added

sprucely (approved):

PR #45, opened 2021-03-18
First crack at the quote page

### LINKED ISSUE #43: Create Quote Page
First pass at the create quote page

srkirkland:
Lots to do still but first crack at what this page might look like.  Fully functional nested add/remove.  Totals need work and prices need to load dynamically (though probably only once).  No ability to save or load an existing quote (though should be fairly easy to add).

I wanted to get this PR up though because so we can discuss if this is a good way to handle the structure, prop passing, etc.  

closes #43

### DIFF
--- a/Harvest.Core/Domain/Project.cs
+++ b/Harvest.Core/Domain/Project.cs
@@ -25,8 +25,10 @@ namespace Harvest.Core.Domain
         [StringLength(200)]
         public string Name { get; set; }
         
-        public int PrincipalInvestigator { get; set; }
+        public int PrincipalInvestigatorId { get; set; }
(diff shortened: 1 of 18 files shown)


jSylvestre:
@srkirkland Got a foreign key error creating a project. Pointing at test db

jSylvestre (approved):

sprucely (approved):
Looks good to me. The async callback defined inside a useEffect pattern is increasingly frowned upon, but there is still not a consensus on how to avoid it outside of third-party async libraries or experimental `Suspense` api. The current pattern is easily understood, so I'm fine with it.

Issue #30, opened 2021-03-12
Add in jquery + validation


srkirkland:
Add jquery and validation into pages that need it.  Probably worth just making as a partial import file since many pages will not need it, but any CSHTML pages with a form will.  Also CSHTML pages with datatables will also need some version of jquery, and they might as well all use the same one.

For reference, a new MVC project from template uses these libs: https://github.com/dotnet/aspnetcore/tree/main/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/wwwroot/lib.

Let's at least load the same versions from CDN, but perhaps at some point we'll want to serve the actual files locally as well for development.

jSylvestre:
https://docs.microsoft.com/en-us/aspnet/ajax/cdn/jquery-validate/

PR #46, opened 2021-03-19
jquery validation cdn

jSylvestre:
Issue #30

### DIFF
--- a/Harvest.Web/Views/Permissions/Create.cshtml
+++ b/Harvest.Web/Views/Permissions/Create.cshtml
@@ -29,3 +29,8 @@
         </form>
     </div>
 </div>
+
+@section Scripts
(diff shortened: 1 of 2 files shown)


sprucely (approved):
Is that really all that's necessary? Do the asp-tags generate the annotations used by jquery-validation?

jSylvestre:
@sprucely Yes, but.... I don't think I have the latest versions, and for Development/Fallback we should have local files. I wasn't sure the best way to get those in the project

srkirkland:
@jSylvestre We can wait on the local fallback, but for that what I would do is just `dotnet create mvc` locally and then steal the file from the official .net template.

srkirkland (approved):

PR #47, opened 2021-03-22
Mostly the access denied page

### LINKED ISSUE #33: Access Denied page
We need an access denied page

jSylvestre:
Closes #33


sprucely (approved):
Looks good. But I think it can be made better...
```
        case 400:
            statusmessage = "That's not how you use a pineapple";
            break;
        case 403:
            statusmessage = "Drop that pineapple";
            break;
        case 404:
            statusmessage = "Where's the pineapple?";
            break;
        default:
            statusmessage = "We need an expert pineappleist";
            break;
```

srkirkland (approved):

PR #48, opened 2021-03-22
Add SlothService and related db changes

sprucely:
This is based loosely on Anlab. It doesn't add anything like an Invoice or Order type to the db; just TransferRequest, which encapsulates multiple Transfer and TransferHistory types. My thought is that once invoices are implemented, a TransferRequest can be generated when an Invoice is approved and processed.


Harvest.Core/Models/SlothModels/TransactionViewModel.cs
```diff
@@ -0,0 +1,37 @@
...
+            TransactionDate = DateTime.UtcNow; //DateTime.UtcNow.ToString(CultureInfo.InvariantCulture);
+            Transfers = new List<TransferViewModel>(2);
+        }
+        public string MerchantTrackingNumber { get; set; }
+
+        public string MerchantTrackingUrl { get; set; }
+
+        public DateTime TransactionDate { get; set; }
+
+        public string Source { get; set; } = "ANLAB Internal Recharge";
```
srkirkland, line 19: TODO for this to change at some point

srkirkland (approved):

Harvest.Core/Services/SlothService.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            var model = new TransactionViewModel
+            {
+                MerchantTrackingNumber = moneyTransfer.Id.ToString(),
+                MerchantTrackingUrl = $"{_slothSettings.MerchantTrackingUrl}/{moneyTransfer.Id}"
+            };
+
+            model.Transfers.Add(new TransferViewModel
+            {
+                Account = moneyTransfer.FromAccount.Number?.ToUpperInvariant(),
```
jSylvestre, line 63: @sprucely If the account number is missing, this would be invalid.

Harvest.Core/Services/SlothService.cs
```diff
@@ -0,0 +1,210 @@
...
+            var token = _slothSettings.ApiKey;
+            var url = _slothSettings.ApiUrl;
+
+            if (string.IsNullOrWhiteSpace(token))
+            {
+                Log.Error("Sloth Token missing");
+            }
+
+            var debit = await _financialService.IsValid(moneyTransfer.FromAccount.Number);
+            var credit = await _financialService.IsValid(moneyTransfer.ToAccount.Number);
```
jSylvestre, line 53: If these debit and credit are invalid, do something with them. Maybe throw an exception?

Harvest.Core/Services/SlothService.cs
jSylvestre: Also, maybe to be consistent, use the debit.KfsAccount like you do for chart and subaccount?

Harvest.Core/Services/SlothService.cs
```diff
@@ -0,0 +1,210 @@
...
+                MerchantTrackingUrl = $"{_slothSettings.MerchantTrackingUrl}/{moneyTransfer.Id}"
+            };
+
+            model.Transfers.Add(new TransferViewModel
+            {
+                Account = moneyTransfer.FromAccount.Number?.ToUpperInvariant(),
+                Amount = moneyTransfer.Amount,
+                Chart = debit.KfsAccount.ChartOfAccountsCode, 
+                SubAccount = debit.KfsAccount.SubAccount,
+                Description = moneyTransfer.Description,
```
jSylvestre, line 67: In other projects I've created a string extension that safely truncates these:

`        public static string SpecialTruncation(this string value, int additionalLength, int maxLength)
        {
            if ((value.Length + additionalLength) <= maxLength)
            {
                return value;
            }

            var max = (value.Length + additionalLength + 3) - maxLength;
            return $"{value.Substring(0, (value.Length - max))}...";
        }`

jSylvestre: Description = $"{order.Project.SpecialTruncation((order.RequestNum.Length + 3), 40)} - {order.RequestNum}",

jSylvestre (approved):
Couple comments, but up to you if you want to change them.

Harvest.Core/Services/SlothService.cs
sprucely: This is equivalent to `SafeToUpper`, how it's done in Anlab... https://github.com/ucdavis/Anlab/blob/1786bab78194fcc618fae52ad18e99b486c1eb18/Anlab.Core/Services/SlothService.cs#L57

Harvest.Core/Services/SlothService.cs
sprucely: I found that function a little challenging to follow, so I split into two, `Truncate` and `TruncateAndAppend`

PR #49, opened 2021-03-24
Jcs/razor l20210324

jSylvestre:


jSylvestre:
If this uses bete10, it gets:
TypeLoadException: Could not load type 'Microsoft.Extensions.Primitives.InplaceStringBuilder' from assembly 'Microsoft.Extensions.Primitives, Version=5.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.

If this uses rc3 (latest prerelease) it gets :
InvalidOperationException: Cannot find reference assembly 'Microsoft.AspNetCore.Antiforgery.dll' file for package Microsoft.AspNetCore.Antiforgery

sprucely:
Harvest.Email is a "Razor Library" project. That will allow emails to be generated even from jobs.

sprucely:
Shouldn't we want to get mjml parsing in there also?

Harvest.Email/Services/EmailBodyService.cs
```diff
@@ -0,0 +1,97 @@
...
+                new TempDataDictionary(
+                    actionContext.HttpContext,
+                    _tempDataProvider),
+                output,
+                new HtmlHelperOptions());
+
+            await view.RenderAsync(viewContext);
+
+            return output.ToString();
+        }
```
sprucely, line 65: I know this is from my commit, but I'm thinking it should be extended to parse the mjml, either automatically or as an option, using [Mjml.AspNetCore](https://www.nuget.org/packages/Mjml.AspNetCore/).

sprucely (changes requested):

Harvest.Email/Services/EmailBodyService.cs
srkirkland: I'm trying to decide if we want to commit to Mjml enough to integrate or if it's better to just pre-compile the mjml into html and stick that in our view.  I'm leaning towards the later now -- i'm not sure why we want the complexity or time of spinning up a node instance to compile a view that is the exact same every time instead of just doing it once and checking that code in.  Perhaps that's something we can build automate if this was a larger project, but for now I'd say let's just stick with what we have and we can add mjml custom rendering if we want it later.

srkirkland (approved):

sprucely (approved):

PR #51, opened 2021-03-29
Sample email.

jSylvestre:
1) Run home/TestBody
2) Paste that into mjml online viewer, copy the html from there
3) Paste that into the cshtm, add in the model stuff at the top
4) replace text with @model. whatever
5) where @media or @import exists, add an extra @ in front


sprucely (approved):

jSylvestre:
Still need to remove mjml stuff out of non mjml views

PR #52, opened 2021-03-29
Add a db project

jSylvestre:


jSylvestre:
This is the schema from test.
The view is just something I have locally

srkirkland (approved):
As long as it is set to not build in any configuration.  I can't tell from the sln file if that's the case but if not you should change it.

jSylvestre:
@srkirkland You just mean this correct?
image

srkirkland:
> @srkirkland You just mean this correct?
> image

Correct -- don't build or deploy in any of the configurations for that project

PR #66, opened 2021-04-08
Basic page for entering expenses

srkirkland:
Should work pretty well except we need to handle the "other" case, but that's a bigger issue that needs to be discussed first.  After this is in, I'll make some tables and hook everything up so we can actually push expenses to the db.


jSylvestre:
The other seemed ok as long as it is in the rate table.

image

jSylvestre (approved):

PR #67, opened 2021-04-09
Archive Rate when price is changed

### LINKED ISSUE #64: Rate values & updating
Proposal:
Rates should never be allowed to change so that we can keep a historical record -- if the value of a rate is updated, we should duplicate the row and then set the old one inactive.  This way we will not only have a good history but expenses can be linked with a FK relationship to the rate and it will always be correct.

jSylvestre:
Closes #64

### DIFF
--- a/Harvest.Web/Controllers/RateController.cs
+++ b/Harvest.Web/Controllers/RateController.cs
@@ -160,16 +160,47 @@ namespace Harvest.Web.Controllers
 
             var user = await _userService.GetCurrentUser();
 
+            //TODO: When the rate is actually used, check the db to see if we need to archive. (If I just created it, and edit the rate, we don't need to archive)
+            var archive = rateToEdit.Price != model.Rate.Price;
 
-            rateToEdit.Account     = model.Rate.Account;
-            rateToEdit.BillingUnit = model.Rate.BillingUnit;
-            rateToEdit.Description = model.Rate.Description;
-            rateToEdit.EffectiveOn = model.Rate.EffectiveOn.FromPacificTime();
-            rateToEdit.Price       = model.Rate.Price;
-            rateToEdit.Type        = model.Rate.Type;
-            rateToEdit.Unit        = model.Rate.Unit;
-            rateToEdit.UpdatedOn   = DateTime.UtcNow;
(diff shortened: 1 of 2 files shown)


Harvest.Web/Controllers/RateController.cs
```diff
@@ -160,16 +160,47 @@ public async Task<ActionResult> Edit(int id, RateEditModel model)
 
             var user = await _userService.GetCurrentUser();
 
+            //TODO: When the rate is actually used, check the db to see if we need to archive. (If I just created it, and edit the rate, we don't need to archive)
+            var archive = rateToEdit.Price != model.Rate.Price;
```
srkirkland, line 164: @jSylvestre actually I think while tracking change of rate itself is probably useful, the real reason to copy is when the account changes.  I'm on the fence about what should happen if they just change the billing unit or description -- but for now the biggest thing would be if the account changes we need to archive.

Harvest.Web/Controllers/RateController.cs
```diff
@@ -160,16 +160,47 @@ public async Task<ActionResult> Edit(int id, RateEditModel model)
...
+            {
+                rateToEdit.Account = model.Rate.Account;
+                rateToEdit.BillingUnit = model.Rate.BillingUnit;
+                rateToEdit.Description = model.Rate.Description;
+                rateToEdit.EffectiveOn = model.Rate.EffectiveOn.FromPacificTime();
+                rateToEdit.Price = model.Rate.Price; //When we check if the rate has been used or not, this may get changed
+                rateToEdit.Type = model.Rate.Type;
+                rateToEdit.Unit = model.Rate.Unit;
+                rateToEdit.UpdatedOn = DateTime.UtcNow;
+                rateToEdit.UpdatedBy = user;
```
srkirkland, line 199: Seems like most of these are the same as above so you can move them out of here and into a common area.  Then do a `if (archive)` and use that to set the createdBy and other stuff that is unique to new entries.

srkirkland (changes requested):

srkirkland (approved):
A little more complicated than I was thinking with that extra method but it does the job and now if we need to add more fields it'll be easy to know where to update them so I think it looks fine

PR #68, opened 2021-04-12
New expense/invoices tables plus save capability

srkirkland:
Can now enter expenses and press save to send to db.  Made new controller as API controller and not SuperController because of API specific routing and anti-forgery issues (similar to peaks but simpler).


jSylvestre (approved):
Locally, changing quantity type on UI to number and TryFloat, was able to save decimals.

PR #71, opened 2021-04-12
Jcs/check if rate used

jSylvestre:


Harvest.Web/Controllers/RateController.cs
```diff
@@ -157,33 +148,59 @@ public async Task<ActionResult> Edit(int id, RateEditModel model)
 
 
             var rateToEdit = await _dbContext.Rates.SingleAsync(a => a.Id == id);
+            var rateUsed = await _dbContext.Expenses.AnyAsync(a => a.RateId == rateToEdit.Id);
```
jSylvestre, line 151: I pulled in main, so the rateUsed was the only actual change

srkirkland (changes requested):
Why bother?  I don't see that it's worth it

jSylvestre:
@srkirkland It prevents cluttering up the table if they edit it and it hasn't been used yet.
But up to you.

srkirkland:
Extra rows in the database wouldn't bother me one bit

jSylvestre:
Not needed

Issue #90, opened 2021-04-16
configure prettier and eslint-prettier


srkirkland:
https://saikat.dev/blog/posts/create-react-app-adding-prettier

PR #92, opened 2021-04-19
officially install and setup prettier eslint rules

srkirkland:
also format all files that don't match.  later we might add in an azure pipelines rule to make sure formatting is correct


srkirkland:
closes #90

PR #94, opened 2021-04-19
quote display 

srkirkland:
Start of quote display, which will be used in quote approval and probably also for historical purposes.  might also be useful for invoicing later.  Stopping at this point though because we need to store more info on the Quote, so i'll do #93 and then come back to this in a future PR

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/QuoteDisplay.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/QuoteDisplay.tsx
@@ -1,0 +1,33 @@
+import React from "react";
+
+import { QuoteContent } from "../types";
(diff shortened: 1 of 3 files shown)


sprucely (approved):
Works for me.
Should the `Approve Quote` button be disabled when project status is set to `PendingAccountApproval`?

jSylvestre (approved):
Works for me.
I'm guessing we can get all the columns formatted later:
image

srkirkland:
> Should the `Approve Quote` button be disabled when project status is set to `PendingAccountApproval`?

Yes, though first we need to define what the status steps are and how to identify them.  But yes that is the goal -- or perhaps the page shouldn't even show if the project isn't in the proper status.

PR #103, opened 2021-04-21
Add Acreage Expense

jSylvestre:
This is using fields that don't exist yet.


Harvest.Core/Services/InvoiceService.cs
```diff
@@ -74,6 +77,38 @@ public async Task<bool> CreateInvoice(int id)
 
         }
 
+        private async Task CreateMonthlyAcreageExpense(Project project)
+        {
+            var amountToCharge = Math.Round((decimal)project.Acres * (project.AcreageRate.Price / 12), 2);
```
jSylvestre, line 82: @srkirkland This is what will be charged per "month" but this value may not match up with the expense Price*Quantity exactly.
1150/12 = 95.8333333...
15 acrers * 95.82 != 1437.50

but, I still end up using the expense Total.

srkirkland (approved):

PR #106, opened 2021-04-21
Add acreage rate selection to quote and update project on approve

### LINKED ISSUE #102: Acreage Rate for project
The Quote needs to select one of the Rates with an Acreage type.
Then that Rate needs to be linked to the project.

I can use that to determine the account I should use when creating the expense when the invoice is created.

image

srkirkland:
Quote can now save additional acreage info and requires choosing an acreage from the rate table.  When that quote is approved, acreage info is set on the project along with a few other details we should have been doing all along.

closes #102


jSylvestre (approved):
🚢

sprucely (approved):

PR #113, opened 2021-04-21
Cyd/create new project

### LINKED ISSUE #83: Field request needs to track crop type (row vs tree)

cydoval:
Closes #83
Screen Shot 2021-04-21 at 3 50 09 PM

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
@@ -41,14 +41,16 @@ export const RequestContainer = () => {
   };
 
   return (
-    <Card>
-      <CardBody>
-        <CardHeader>Create Field Request</CardHeader>
-        <Container>
-          <Row>
-            <Col>
-              <FormGroup>
-                <Label>When to Start?</Label>
+    <div className="card-wrapper card-medium">
+      <div className="card-content">
+        <div className="card-head">
+          <h2>Create Field Request</h2>
+        </div>
+        <div className="row">
+          <div className="col-md-6">
+            <div className="form-group">
+              <Label>When to Start?</Label>
+              <div className="input-group">
                 <DatePicker
                   format="MM/dd/yyyy"
                   required={true}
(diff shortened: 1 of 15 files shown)


sprucely (approved):
Looks good to me. I'm assuming you converted to using bootstrap classes directly because it gives you more control?

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
```diff
@@ -72,60 +80,78 @@ export const RequestContainer = () => {
...
+          <Label>Which type of crop will we grow?</Label>
+          <div className="custom-control custom-radio">
+            <input
+              type="radio"
+              id="customRadio1"
+              name="customRadio"
+              className="custom-control-input"
+              style={{ zIndex: 1 }} //prevent class custom-control-input from blocking mouse clicks
+              value="Row"
+              checked={project.cropType === undefined || project.cropType === "Row"}
```
srkirkland, line 97: @sprucely could be nice to just default it to row and make it not able to be undefined.

srkirkland (approved):

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
srkirkland: @sprucely looking at the code i don't really see how cropType could be undefined since it's basically a two value enum?

jSylvestre:
This still happens:
image

sprucely:
> 
> 
> This still happens:
Forgot about that. I'll try to fix it...

srkirkland:
@sprucely if you can't fix it easily have Cal look at it, he's fixed it before in our other projects that use the same component

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
sprucely: The interface/type doesn't allow undefined, but state was initialized from an almost empty object. I'll just make the initial state include `cropType: "Row"`

jSylvestre (approved):

PR #114, opened 2021-04-22
First invoice 30 days after project starts

jSylvestre:


Harvest.Core/Services/InvoiceService.cs
```diff
@@ -42,12 +42,13 @@ public async Task<bool> CreateInvoice(int id)
             }
 
             var now = DateTime.UtcNow;
-            if (project.Start > now)
+            if (project.Start.AddDays(30) > now) //Start doing invoices 30 days after the project starts
```
sprucely, line 45: Should AddMonth(1) be used instead? It takes into account the number of days in a month, so Jan 31 + 1 month = Feb 28 (or Feb 29 on leap year)

Harvest.Core/Services/InvoiceService.cs
```diff
@@ -42,12 +42,13 @@ public async Task<bool> CreateInvoice(int id)
             }
 
             var now = DateTime.UtcNow;
-            if (project.Start > now)
+            if (project.Start.AddDays(30) > now) //Start doing invoices 30 days after the project starts
             {
                 //Project hasn't started yet. (Is Start UTC? if not, it should be)
```
srkirkland, line 47: (I believe start is UTC but we only let them choose date so it doesn't really make a difference)

srkirkland (approved):

Harvest.Core/Services/InvoiceService.cs
srkirkland: Good idea, doesn't make much difference but I think it's a good idea to be careful on date math

PR #118, opened 2021-04-22
Create change request from existing project

### LINKED ISSUE #112: Create approved flag in Projects table
Discussion: How should we handle change requests?  I'm thinking the simplest thing is to add a field to the projects table for like ChangeRequest (TXT) where they describe what they want changed, and we also update the status to something like `ChangeRequested` to indicate the difference.  Then when a new change quote is submitted, we go back to `PendingQuoteApproval` status.  The downside is that we want to keep taking expenses the whole time, and the quote might never be approved.  So I'm thinking the simplest thing is to add an `Approved` flag which when true means that it's a valid project that can take expenses.  We would still need to determine how to handle proposed accounts though. Perhaps we can do something clever with merging multiple projects (as in, should the change request be a separate project which merges back to the main one once accepted?)

srkirkland:
At Request/Create/[originalProjectId] then it'll load up the request form with the values from that original project and allow you to create a change request.  This will then get created with the changes and proper additional originalProjectId.  Name is recreated, which we may or may not want.   Closes #112

### DIFF
--- a/Harvest.Core/Domain/Project.cs
+++ b/Harvest.Core/Domain/Project.cs
@@ -13,15 +13,15 @@ namespace Harvest.Core.Domain
     {
         [Key]
         public int Id { get; set; }
-        
+
         public DateTime Start { get; set; }
-        
+
         public DateTime End { get; set; }
-        
+
         // TODO: make this biggere in case we have a bunch of crops
         [StringLength(50)]
         public string Crop { get; set; }
-        
+
         [MaxLength]
         public string Requirements { get; set; }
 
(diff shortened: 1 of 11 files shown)


Harvest.Core/Domain/Project.cs
```diff
@@ -114,13 +122,13 @@ internal static void OnModelCreating(ModelBuilder modelBuilder)
...
                 .OnDelete(DeleteBehavior.Restrict);
-
         }
         public class Statuses
         {
             public const string Requested = "Requested";
             public const string PendingAccountApproval = "Pending Account Approval";
-            public const string Active    = "Active";
+            public const string Active = "Active";
+            public const string ChangeRequested = "ChangeRequested";
```
jSylvestre, line 131: Add a space in the string?
Also, add it to the TypeList? (Not a big deal as I don't think we are using that yet if ever.)

jSylvestre (approved):
Should you be able to change the PI?
Should there be a text field specific to the change request? (And show that on the quote page if there is anything in it)
If you change the PI (and maybe the dates) the project name changes. I don't think this should happen.
Other than the project name, there should be something that indicates which project (original) that will change.

Harvest.Core/Domain/Project.cs
srkirkland: I don't like spaces in the string, makes the db match harder too.  but I should probably change the others so it's consistent

srkirkland:
> Should you be able to change the PI?
> Should there be a text field specific to the change request? (And show that on the quote page if there is anything in it)
> If you change the PI (and maybe the dates) the project name changes. I don't think this should happen.
> Other than the project name, there should be something that indicates which project (original) that will change.

We can copy the name from the original project if we want, but i could see reasons why you would change the PI.  The thing indicating which project will change is the OriginalProjectId column

Issue #110, opened 2021-04-21
Email Notification Templates


srkirkland:
Need some email templates created.  For now they can be simple but should convey basic info. Follow the mjml example in the Harvest.Email lib and the homeController.  If you have some good thoughts about how to create a shared header/body for the emails that'd be nice too.  Payments emails might be a good example if looking for styles

Farm Manager:
* [x] 1. New request comes in
* [x] 2. Quote is accepted
* [x] 3. Quote is rejected
* [x] 4. Change request is created

Professor:
* [x] 1. Quote ready for approval

jSylvestre:
Don't forget to use mailtrap. Much easier to check against

jSylvestre:
This is the current shared header:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Email/Views/Shared/_EmailLayout_mjml.cshtml

PR #123, opened 2021-04-23
Quote Decision Emails

jSylvestre:
Issue #110 
https://mailtrap.io/inboxes/1276099/messages/2170492731
https://mailtrap.io/inboxes/1276099/messages/2170492870

### DIFF
--- a/Harvest.Web/Services/EmailService.cs
+++ b/Harvest.Web/Services/EmailService.cs
@@ -17,6 +17,9 @@ namespace Harvest.Web.Services
     {
         Task<bool> ProfessorQuoteReady(Project project);
         Task<bool> NewFieldRequest(Project project);
+
+        Task<bool> QuoteApproved(Project project);
+        Task<bool> QuoteDenied(Project project);
     }
 
     public class EmailService : IEmailService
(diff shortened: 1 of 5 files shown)


Harvest.Web/Services/EmailService.cs
```diff
@@ -64,10 +67,14 @@ public async Task<bool> ProfessorQuoteReady(Project project)
 
         }
 
+        private async Task<string[]> FieldWorkersEmails()
+        {
+            return await _dbContext.Permissions.Where(a => a.Role.Name == Role.Codes.FieldManager).Select(a => a.User.Email).ToArrayAsync();
+        }
```
sprucely, line 73: Shouldn't this be filtering on `Roles.Codes.Worker`?

sprucely (approved):
Looks good as long as that filter on `Role.Codes.FieldManager` is correct.

Harvest.Web/Services/EmailService.cs
jSylvestre: Not worker, But I can change the name of the method in my next pr

PR #126, opened 2021-04-23
Replace JsonConvert with JsonSerializer

### LINKED ISSUE #107: Replace uses of JsonConvert with JsonSerializer

sprucely:
Closes #107 

I tested several Serialize/Deserialize operations and didn't see any problems.

### DIFF
--- a/Harvest.Jobs.Invoice/Program.cs
+++ b/Harvest.Jobs.Invoice/Program.cs
@@ -60,7 +60,7 @@ namespace Harvest.Jobs.Invoice
             services.AddScoped<ISlothService, SlothService>();
             services.AddScoped<IFinancialService, FinancialService>();
             //services.Configure<SparkpostSettings>(Configuration.GetSection("Sparkpost"));
-            services.AddScoped(provder => StandardJsonOptions.GetOptions());
+            services.AddSingleton(provder => StandardJsonOptions.GetOptions());
 
             return services.BuildServiceProvider();
         }
--- a/Harvest.Jobs.Invoice/Program.cs
+++ b/Harvest.Jobs.Invoice/Program.cs
@@ -3,6 +3,7 @@ using System.Threading.Tasks;
 using Harvest.Core.Data;
 using Harvest.Core.Models.Settings;
 using Harvest.Core.Services;
+using Harvest.Core.Utilities;
 using Harvest.Jobs.Core;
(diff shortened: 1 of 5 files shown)


srkirkland (approved):
👍🚀

jSylvestre (changes requested):
Doesn't work.
https://localhost:44308/Rate/Details/3

image

vs.

image

sprucely:
I added some significant changes, so I thought it could use another review.

Harvest.Jobs.Invoice/Program.cs
```diff
@@ -59,6 +60,7 @@ private static ServiceProvider ConfigureServices()
             services.AddScoped<ISlothService, SlothService>();
             services.AddScoped<IFinancialService, FinancialService>();
             //services.Configure<SparkpostSettings>(Configuration.GetSection("Sparkpost"));
+            services.AddScoped(provder => StandardJsonOptions.GetOptions());
```
srkirkland, line 63: To reignite this debate for fun, why scoped?  it's this really just a singleton?  The options never changes or needs to be re-allocated

Harvest.Web/Startup.cs
```diff
@@ -52,7 +55,15 @@ public void ConfigureServices(IServiceCollection services)
             services.AddControllersWithViews(options =>
             {
                 options.Filters.Add<SerilogControllerActionFilter>();
-            }).AddJsonOptions(options => options.JsonSerializerOptions.Converters.Add(new NetTopologySuite.IO.Converters.GeoJsonConverterFactory()));
+            }).AddJsonOptions(options =>
```
jSylvestre, line 58: @sprucely If we have these options here in startup, do we still need them to be specified when we deserialize?

jSylvestre (approved):

Harvest.Web/Startup.cs
sprucely: I believe so, because these options apply only to the automatic serialization/deserialization performed by aspnetcore in the request/response pipeline.

Harvest.Jobs.Invoice/Program.cs
sprucely: Yeah, I was just doing it because the services it was going into are scoped. I'll change it to singleton...

PR #128, opened 2021-04-23
Add request form valdiation

### LINKED ISSUE #89: Validate request create page
Add yup nuget package (just like in CECDSS project) and validate the project is valid in RequestContainer.tsx.  For now we'll just wait until the create button has been pressed and then do validation and show any errors in a list right above the create button.

wongband:
resolve #89


jSylvestre (approved):
FYI, it would be nice to have the errors under the fields they are triggered for, also, the date's error will not make sense to the end user. But looks good.

image

Harvest.Web/ClientApp/src/types.ts
```diff
@@ -181,3 +181,24 @@ export interface ProjectAccount {
...
+  id: number;
+  start: string;
+  end: string;
+  crop: string;
+  cropType: string;
+  requirements?: string;
+  principalInvestigator: PrincipalInvestigator;
+}
+
+export interface PrincipalInvestigator {
```
srkirkland, line 195: @wongband this looks a lot like the `User` type from this file, which has everything except nameAndEmail which isn't really something they input or use so I don't think it's something that needs to be error checked.  so maybe you can just use that?  I think eventually we'll want to also replace RequestInput with some subclass of `Project` but that will take more work so you can leave that part as-is

srkirkland (commented):
Agree with Jason's comments but I think those are things we can come back to -- getting some basic validation in is a good start.  When you start validating the expenses page it's going to be more important to display errors next to the field that has an error because of the repeating nature of the components.

Harvest.Web/ClientApp/src/types.ts
wongband: I created PrincipalInvestigator since nameAndEmail seems to be an input when the button is submitted. This is what we get when the user tries to create a request:

```json
{
"id":0,"firstName":"Scott",
"lastName":"Aldinger",
"email":"saldinger@ucdavis.edu",
"iam":"1000153077",
"kerberos":"scott",
"name":"Scott Aldinger",
"nameAndEmail":"Scott Aldinger (saldinger@ucdavis.edu)"
}
```

srkirkland: IIRC It's just ignored once sent to the server so we don't need to validate.  But i guess it's not a problem since they won't have control to mess up the values anyway.  I'm fine if you leave it as is if it's not a quick change, I don't really remember how that value gets there without looking more.

wongband: Okay, so I changed it so we don't have to create a whole new type.

srkirkland:
@wongband looks like you have a merge conflict but once you resolve it i think this is good to go

PR #136, opened 2021-04-27
Added error checking for emtpy quantities in expenses

### LINKED ISSUE #80: Expenses with zero amounts
Either require an amount, or when the expenses are saved, skip ones that don't have a quantity

wongband:
resolve #80


srkirkland (changes requested):
Since this is a list of expenses I don't think it'll do having a generic alert, we'll need to show the error associated with the row where the error is.  So perhaps each expense can keep track of it's validity and show the message.

wongband:
Yea I can make it so that an error is displayed below the expense row.

jSylvestre (changes requested):
Not working correctly.

image

jSylvestre (approved):
Minor issues can be fixed later.
🚢

PR #138, opened 2021-04-27
Bring in react-table and show Projects in react-table

srkirkland:
Brings in react-table and all helpers.  You'll need to npm install.


jSylvestre (approved):
Shinny.
How difficult to have an indicator of column sort?

srkirkland:
> Shinny.
> How difficult to have an indicator of column sort?

Easy peasy, columns can just be a React component so we can have it be whatever we want

PR #139, opened 2021-04-28
Audit fix

jSylvestre:
Ran npm audit fix


srkirkland (approved):
strange fix that just deletes some lines but i don't know much about the package-lock so if it works then ok

PR #141, opened 2021-04-28
Create project details in React and refactor some common headers

srkirkland:
Eventually we'll want to move around our API calls so we don't end up with so many manual react server methods but for now this is working pretty well.


Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
```diff
@@ -16,9 +17,9 @@ export const ProjectTable = (props: Props) => {
       {
         Cell: (data: Cell<Project>) => (
           <div>
-            <a href={`/project/details/${data.row.original.id}`}>
+            <Link to={`/project/details/${data.row.original.id}`}>
               #{data.row.original.id}
-            </a>
+            </Link>
             <p>{data.row.original.name}</p>
```
jSylvestre, line 23: @srkirkland Move this into the link name?

jSylvestre (approved):

sprucely (approved):
Looks good to me.

PR #144, opened 2021-04-28
Create a ProjectHistoryService

### LINKED ISSUE #115: Project history
Create a history service, similar to PEAKS or other apps.  Actions again the project should be recorded in the history table, definitely every time the project status changes but also on other changes. ex:

ProjectRequested
Quote created/updated/submitted/approved
Accounts approved
Change requested

sprucely:
Closes #115


Harvest.Core/Harvest.Core.csproj
```diff
@@ -1,9 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
```
srkirkland, line 4: I'm really uncomfortable with this change, though I understand why you wanted to do it.  But I'm not sure why I don't like it or if there would be any downsides.  What do you think?

sprucely: That seemed like it was the only way I could get access to `IHttpContextAccessor`. It kept saying something like "Wrong project type" And even then, I was having trouble. Adding `<FrameworkReference Include="Microsoft.AspNetCore.App" />` did the trick.

sprucely: I can't think of any downsides. The only projects referencing it will be under our control and with a compatible version. And any third-parties would be accessing it via API.

jSylvestre: I had the same issue when I was trying to include the email service. Ended up not doing it.

srkirkland: Ok well then we can go with the change, I can't see a scenario in which we would need to use this core lib from a different project framework.  And then we could move the email service in which is useful too.

Harvest.Core/Services/ProjectHistoryService.cs
```diff
@@ -0,0 +1,51 @@
...
+        Task<ProjectHistory> AddProjectHistory(Project project, string action, string description, object detailsToSerialize, User user = null);
+    }
+
+    public class ProjectHistoryService : IProjectHistoryService
+    {
+        private readonly IUserService _userService;
+        private readonly AppDbContext _dbContext;
+        private readonly JsonSerializerOptions _jsonOptions;
+
+        public ProjectHistoryService(IUserService userService, AppDbContext dbContext, JsonSerializerOptions jsonOptions)
```
srkirkland, line 23: Ooh this is nice -- i'm guessing that this is injected based on the defaults you pass to the mvc/controllers method AddJsonOptions?

srkirkland (approved):

Harvest.Core/Services/ProjectHistoryService.cs
sprucely: Not from `AddJsonOptions`, but `services.AddScoped(provder => StandardJsonOptions.GetOptions());`. I believe AddJsonOptions only configures for payloads being passed to/from controller actions. But the options for all of them come from `StandardJsonOptions.GetOptions`.

PR #153, opened 2021-05-03
Move GeoJson options to Harvest.Core

sprucely:
This will allow GeoJson (de)serialization to be performed in jobs.

I wasn't sure if GeoJson should be part of the standard `JsonSerializerOptions`, so I left it to be added as needed via extension method `JsonSerializerOptions.WithGeoJson()`


Harvest.Core/Harvest.Core.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="5.0.1" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="5.0.1" />
     <PackageReference Include="NetTopologySuite" Version="2.1.0" />
```
jSylvestre, line 16: @sprucely Does NetTopologySuite need the latest stable (2.20) for the one you added below?

jSylvestre (approved):

Harvest.Core/Harvest.Core.csproj
sprucely: It probably wouldn't hurt. But I think I'll do it in another PR when we're updating a bunch of dependencies.

PR #156, opened 2021-05-06
Table design for Project/rates/permissions

cydoval:


sprucely (commented):
Here's how it looks for me. There are some pages that have the datatable adjusting its width, but this one doesn't seem to be doing that. I also vaguely recall some sort of datatables config or plugin that would prioritize collapsing certain columns based on available space.

image

Harvest.Web/ClientApp/src/Projects/ProjectListContainer.tsx
```diff
@@ -22,7 +22,17 @@ export const ProjectListContainer = () => {
 
   return (
     <div>
-      <Link to="/request/create">Create New</Link>
+      <div className="row justify-content-between mb-3">
+        <div className="col">
+          <h1>Projects</h1>
+        </div>
+        <div className="col text-right">
+          <a className="btn btn-sm btn-link " href="/request/create">
```
srkirkland, line 30: Unfortunately you can't replace `<Link>` with `<a>` because they are not really the same thing so we'll need to put that back.  But it works just like <a> so you can put content inside and add classes to it.

srkirkland (changes requested):

jSylvestre (approved):

sprucely (approved):

PR #163, opened 2021-05-11
Remove all things mobile

sprucely:
Not sure why Visual Studio insists on regenerating project guids any time it modifies the sln file. But it doesn't seem to hurt anything.


sprucely:
Keep meaning and subsequently forgetting to revert those types of changes.

PR #164, opened 2021-05-12
Process invoices on first monday each month

### LINKED ISSUE #161: Invoice should be created/processed on 1st business day of the month

sprucely:
Yep, it really is just a cron update: `0 0 13 1-7 * Mon` That translates to "1pm every day-of-month that falls within 1 through 7 and is a Monday".

Closes #161

Edit: Modified it to run every week day of month and the InvoiceService will determine if a given active project needs an invoice to be created on a given month. Wasn't sure if every day was appropriate or if it should just rerun for the first week. I don't see any harm either way.

### DIFF
--- a/Harvest.Core/Services/InvoiceService.cs
+++ b/Harvest.Core/Services/InvoiceService.cs
@@ -31,6 +31,15 @@ namespace Harvest.Core.Services
         }
         public async Task<bool> CreateInvoice(int id)
         {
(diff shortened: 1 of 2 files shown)


srkirkland (approved):
This approach seems solid to me 👍

jSylvestre (approved):
Seems good to me...

PR #166, opened 2021-05-12
Cyd/quote creation

cydoval:
Screen Shot 2021-05-12 at 1 40 01 PM

### DIFF
--- a/Harvest.Web/Views/Permissions/Index.cshtml
+++ b/Harvest.Web/Views/Permissions/Index.cshtml
@@ -9,7 +9,7 @@
 <h1>Users with roles</h1>
     </div>
     <div class="col text-right">
- <a class="btn btn-sm btn-link" asp-action="Create"><i class="fas fa-plus fa-xs"/></i> Add new User/Role Font Awesome here</a>
+        <a class="btn btn-sm btn-link" asp-action="Create"><i class="fas fa-plus fa-xs"/></i> Add new User/Role Font Awesome here</a>
     </div> 
     </div>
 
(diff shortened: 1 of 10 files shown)


cydoval:
@srkirkland I figure we need a way to remove an activity, so i've added an A tag line 70 on activityform.tsx that can be our trigger for removal

Harvest.Web/Views/Permissions/Index.cshtml
```diff
@@ -9,7 +9,7 @@
 <h1>Users with roles</h1>
     </div>
     <div class="col text-right">
- <a class="btn btn-sm btn-link" asp-action="Create"><i class="fas fa-plus fa-xs"/></i> Add new User/Role Font Awesome here</a>
+        <a class="btn btn-sm btn-link" asp-action="Create"><i class="fas fa-plus fa-xs"/></i> Add new User/Role Font Awesome here</a>
```
sprucely, line 12: Do we still need the "Font Awesome here" here?

sprucely (approved):

Harvest.Web/Views/Permissions/Index.cshtml
srkirkland: @sprucely  IIRC we are not loading FontAwesome on the CSHTML pages, just via react, so we still do need to figure out how to add font awesome here.  Likely we'll want to add it to the styles section on just the pages that it's needed and not the layout page so it is not double-loaded.

srkirkland (approved):

PR #169, opened 2021-05-14
Enable deleting activities

sprucely:
Double-checked, and quote totals are getting updated when an activity is deleted.

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/ActivitiesContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/ActivitiesContainer.tsx
@@ -25,6 +25,11 @@ export const ActivitiesContainer = (props: Props) => {
 
     props.updateQuote({ ...props.quote, activities: [...allActivities] });
   };
+  const deleteActivity = (activity: Activity) => {
+    const allActivities = props.quote.activities.filter((a) => a.id !== activity.id);
+    props.updateQuote({ ...props.quote, activities: [...allActivities] });
+  }
+
   return (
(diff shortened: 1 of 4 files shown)


jSylvestre (approved):
Might be nice that if all the stuff in an activity is removed, that the activity gets removed when it saves. Or error validation.

srkirkland (approved):

sprucely:
@srkirkland @jSylvestre Do you see any issues with my latest commit?

srkirkland:
@sprucely What is the point of the clone? Just in case the save fails?  I think the idea of removing 0 valued stuff is fine but I'm not sure the point of the rest.

sprucely:
Yeah, I don't like modifying state just before an operation that can fail.

sprucely:
I also like cloning state that might be modified in a modal that can be cancelled.

srkirkland:
If we are just changing the location on save (and really should just do a react router redirect in the future, why save the state at all? We are just going to immediately dispose the component.

sprucely:
I thought the window.location thing might be replaced with something that holds on to the state.

sprucely:
Do you want me to remove the clone stuff?

srkirkland:
I don't think so. It'll be replaced but the state shouldn't be retained since it'll be a different route and this should swap out the top level component and remove the state long with everything else in the component.

srkirkland:
> Do you want me to remove the clone stuff?

Yea. I think it was nice and clean before and we should leave it like that for now.

PR #177, opened 2021-05-18
Add ability to specify description on certain expense types

srkirkland:


Harvest.Web/ClientApp/src/Quotes/WorkItemsForm.tsx
```diff
@@ -38,22 +38,30 @@ export const WorkItemsForm = (props: Props) => {
...
         rate: rate.price,
+        unit: rate.unit,
         total: 0,
       });
     }
   };
 
+  // TODO: Determine a better way of working out which other options need extra description text
+  const requiresCustomDescription = (unit: string) => {
+    return props.category === "other" && unit === "Unit";
```
jSylvestre, line 54: @srkirkland Other not other ?

srkirkland: @jSylvestre has to be other lowercase to work.  Though it's just an optimization we don't even need that part of the check. but i'll leave for now until we determine a better way to detect these cases.

jSylvestre (approved):

PR #182, opened 2021-05-19
Allow manual invoice creation in InvoiceService

### LINKED ISSUE #170: Add ability to manually create invoice instead of waiting until job does it
Should use the same logic as the job.  ADMIN only feature for now.

sprucely:
Closes #170

### DIFF
--- a/Harvest.Core/Models/Result.cs
+++ b/Harvest.Core/Models/Result.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Harvest.Core.Models
+{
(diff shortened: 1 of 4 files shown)


jSylvestre (approved):
Didn't test, but looks good.

srkirkland (changes requested):
I can't say exactly how, but see if you can clean it up so you aren't just adding `!manualOverride` repetitively to all the if statements, which I think is just making things a little less clear.  Perhaps you could do a `if (!manualOverride)` and use that to skip over all the checks instead of one at time?

sprucely:
I shifted some things around a bit. Looks cleaner. There's a round trip to fetch project that could have been avoided the in previous version, but that's the only logical difference.

Issue #159, opened 2021-05-11
Don't create/process invoice if total project paid $$ is over quoted amount


srkirkland:
Need to send regular notification emails to farm managers that a new quote/change request is needed and project won't be billed until a higher amount is agreed upon.

PR #184, opened 2021-05-19
Jcs/email when expense goes over quote

jSylvestre:
https://mailtrap.io/inboxes/1276099/messages/2205081936

image

part of issue #159


jSylvestre:
I don't think this can be called from the invoice service directly because it is in core...
Would need to change the job and examine the error

srkirkland (approved):

PR #189, opened 2021-05-28
Add invoice detail page

### LINKED ISSUE #172: Invoice detail
make a nice invoice detail page -- likely via react.  Should look something like this: https://www.sketch.com/s/2039549e-040d-41c3-89ca-c2d3d4d344df/a/YG5nZqv except for the bottom part about quotes and such.  The existing "approve quote" screen should already contain almost all of the needed components, re-use them whenever possible, ex: https://localhost:5001/request/approve/5.

sprucely:
I removed the `ReferenceHandler.Preserve` json option, since it seems to cause more headaches than it solves.

Also made rate type capitalization consistent and enforced for future code by changing it's type from `string` to a union of `"Acreage" | "Equipment" | "Labor" | "Other"`. After merging, some data in `Expenses.Type` and `Type` property of `WorkItems` in the `Quotes` table will need to be corrected.

Closes #172


Harvest.Web/Controllers/TempInvoiceController.cs
```diff
@@ -12,13 +12,13 @@
...
     [Authorize]
-    public class InvoiceController : SuperController
+    public class TempInvoiceController : SuperController
     {
         private readonly AppDbContext _dbContext;
         private readonly IUserService _userService;
         private readonly IInvoiceService _invoiceService;
 
-        public InvoiceController(AppDbContext dbContext, IUserService userService, IInvoiceService invoiceService)
+        public TempInvoiceController(AppDbContext dbContext, IUserService userService, IInvoiceService invoiceService)
```
jSylvestre, line 21: @sprucely The plan is to replace this when the other methods like unbilled are moved into react?

jSylvestre (approved):

Harvest.Web/Controllers/TempInvoiceController.cs
jSylvestre: Should have said remove, not replace.

sprucely: Not sure. I was just getting it out of the way.

PR #192, opened 2021-06-01
Added modal and edit button to popup

### LINKED ISSUE #185: Allow editing of field detail in quote map
If the field needs to change location we don't need to support that yet, unless we can figure out an easy way. But I would like to allow editing field details at least, like changing name or description. 

When a field is created you can click on it and there is a pop up with details and a remove button. The edit can go there and should just bring the modal back up.

wongband:
resolve #185


jSylvestre (approved):
Works ok. Should probably have some instructions here for how to do it, 
image

Maybe: To edit a field, click on it in the map and choose either edit or remove.

Also, a spacer of some kind would be nice. But that is minor:
image

sprucely (approved):
Looks okay. I was going to look into using some type of modal hook to make opening/updating a little less convoluted. But then I noticed the parent is a class component. ¯\\_(ツ)_/¯

PR #200, opened 2021-06-02
accounts are auto approved when quote is approved

### LINKED ISSUE #158: Remove account approval step once PI approved quote

srkirkland:
closes #158


sprucely (approved):
Looks good. Is there a guideline about html formatting? Keep lines short? New line for each attribute when there is more than one?

srkirkland:
> Looks good. Is there a guideline about html formatting? Keep lines short? New line for each attribute when there is more than one?

@sprucely we don't really have a guideline but in this case it wasn't a formatting problem it was a nesting problem.  We had an `<li>` nested inside an `<li>` just because I think someone typed in wrong, and we also had a `<a>` link when it should have been a `<button>` (for accessibility).

jSylvestre:
@srkirkland Should the approve quote list already entered accounts?

jSylvestre (approved):

PR #202, opened 2021-06-02
Fix invoice

sprucely:
A few tweaks to get things serializing and rates displaying.


jSylvestre (approved):
Works for me, but I really don't like having to just through all these hoops when newtonsoft worked fine.

srkirkland (approved):

PR #221, opened 2021-06-04
No trailing -

jSylvestre:


sprucely (commented):
What about if subaccount is blank and account is not?

jSylvestre:
> What about if subaccount is blank and project. is not? Is "--{Project}" valid? [edit] I meant Project, not account.

Then we want the extra dash. Otherwise we may confuse subaccount and project
But.... Project even if supplied is currently being ignored

srkirkland (approved):

sprucely (approved):

PR #224, opened 2021-06-05
Swe/validation

sprucely:
Here's a proof-of-concept using Zod and react-use-form-state. The only annoying thing I've found so far with Zod is that transformations occur after parsing, which means you can't parse a numeric string to a number type. One would have to do something even more janky than my workaround like define the target property as a number | string union to allow such. That's why you'll see a few `parseInt`s and `Number()`'s in the form.

### DIFF
--- a/Harvest.Web/ClientApp/src/types.ts
+++ b/Harvest.Web/ClientApp/src/types.ts
@@ -1,3 +1,5 @@
+import { z } from "zod";
+
 export type CropType = "Row" | "Tree"
 export type RateType = "Acreage" | "Equipment" | "Labor" | "Other"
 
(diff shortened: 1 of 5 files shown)


Harvest.Web/ClientApp/src/types.ts
```diff
@@ -171,17 +173,20 @@ export class WorkItemImpl implements WorkItem {
...
-}
+
+export const WorkItemSchema = z.object({
+  id: z.number().int().default(0),
+  activityId: z.number().int(),
+  type: z.union([z.literal("Acreage"), z.literal("Equipment"), z.literal("Labor"), z.literal("Other")]),
+  rateId: z.number().int().default(0),
+  rate: z.number().min(0),
+  description: z.string(),
+  quantity: z.number().min(0),
```
jSylvestre, line 184: @sprucely instead of min(0), probably just .positive()

PR #225, opened 2021-06-07
Disabled quote approval button if account number is not 100%

### LINKED ISSUE #199: project approval validation
When approving a quote, ex: https://localhost:5001/request/approve/23, the Approval button should only be able to be clicked if there is at least one account selected and all acct percentages add up to 100%.  Otherwise there should be an error message.

wongband:
resolve #199

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
@@ -1,6 +1,6 @@
 // typeahead input box that allows entering valid account numbers
 // each one entered is displayed on a new line where percentages can be set
-import React, { createRef, useState } from "react";
+import React, { createRef, useEffect, useState } from "react";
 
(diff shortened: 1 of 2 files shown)


jSylvestre (approved):

Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
```diff
@@ -22,6 +23,16 @@ export const AccountsInput = (props: Props) => {
...
 
   const { accounts, setAccounts } = props;
+
+  useEffect(() => {
+    let total = 0;
+    for (let i = 0; i < accounts.length; i++) {
+      total += accounts[i].percentage;
+    }
+
+    total == 100 ? props.setDisabled(false) : props.setDisabled(true);
```
srkirkland, line 33: just for fun -- I like doing `props.setDisabled(total !== 100)` in these cases instead of using the ternary operator since it is telling you when to set disabled without having to read the ? : logic out

PR #226, opened 2021-06-08
User name instead of login id

### LINKED ISSUE #206: Show user full name instead of loginID
Just copy the way I did it in PEAKS

https://github.com/ucdavis/Peaks/blob/master/Keas.Mvc/Extensions/ClaimsExtensions.cs.  We want to show it on the layout page.

jSylvestre:
resolves #206

image


srkirkland (changes requested):
Add `Harvest.Web.Extensions` to the `_ViewImports` file instead of on each page you want to use it.

srkirkland (approved):

PR #228, opened 2021-06-09
Cyd/sitecleanup2

cydoval:
numerous changes including activity page and project details page

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/ActivityForm.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/ActivityForm.tsx
@@ -3,6 +3,8 @@ import React from "react";
 import { Input } from "reactstrap";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faMinusCircle } from "@fortawesome/free-solid-svg-icons";
+import { faCopy } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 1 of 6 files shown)


Harvest.Web/ClientApp/src/Quotes/ActivityForm.tsx
```diff
@@ -56,8 +58,54 @@ export const ActivityForm = (props: Props) => {
...
-            <label> Activity Name</label>
+          <div className="col-md-12">
+            <div className="row justify-content-between">
+              <div className="col-md-6">
+                <label> Activity Name</label>
+              </div>
+
+              <div className="col-md-6 text-right">
+                <button className="btn btn-link btn-sm">
+                  Adjust Year <FontAwesomeIcon icon={faCalendarWeek} />
```
srkirkland, line 69: Do we still need adjust year if we have that dropdown toggle?

srkirkland (approved):

Harvest.Web/ClientApp/src/Quotes/ActivityForm.tsx
cydoval: I left that there as a reference to how i want it to look once the other one is hooked up

jSylvestre (approved):

PR #230, opened 2021-06-09
Labels for field request crop type selection to be clickable

### LINKED ISSUE #193: Create field request crop type selection
Clicking the text should allow you to change the crop type, not just clicking the tiny radio button.

wongband:
resolve #193


srkirkland (approved):
@wongband while you are in there can you change the radio inputs to have names/ids that match better?  like say `rowCropInput` and `treeCropInput`?  Then change your update of htmlFor so everything is hooked up nicely

wongband:
@srkirkland I can do that

PR #231, opened 2021-06-11
Form validation with yup and useFormik

sprucely:
I based this off PR #224. Couldn't get quite as far, because Formik wants to manage the whole form and not just small parts of it. The only place you can get your data back out of the form is via `onSubmit()` and perhaps from the promise returned by `submitForm()`. The return type of `useFormik` is not explicit, but I found a way to extract it and declare my own type in order to pass it around. I think it's doable, and a little cleaner than react-use-form-state, if we're willing to have one `useFormik` per form and get the updated object all at once rather than the current pattern of propagating data via callback props.


srkirkland (commented):
The code isn't bad at all but I don't get how it would work.  The updateWorkItems() is never being called so how will the parent container ever get the real value and be able to submit?

sprucely:
@srkirkland That's what I meant regarding Formik wanting to manage the whole form. There isn't a form-level event that will allow us to get data out of the WorkItem form as it is being changed. So instead of calling useFormik from within WorkItemForm, it should be called once from QuoteContainer and passed to the nested components. That way the entire QuoteContent object can be retrieved from formik at once when a user clicks Submit. If we wanted to keep the structure somewhat similar to what we currently have, we would have to do something like creating a subform wrapper or replacing updateWorkItems() with a submitRequested() function. I have seen examples of both, and they are kinda hacky.

PR #232, opened 2021-06-14
Implemented change accounts feature

### LINKED ISSUE #217: Change accounts for any active project:  PI and FM + SUP
@wongband work off this branch https://github.com/ucdavis/Harvest/tree/srk/change-accounts.  I created the shell of the change account page on it -- you can go to https://localhost:5001/request/changeAccount/26 or change the number at the end to any sample project you are using.

I put instructions in a TODO comment in AccountChangeContainer.tsx.  Basically you've seen the page where they specify accounts and then approve a quote.  This new page lets them change those accounts at any time.

wongband:
resolve #217

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
@@ -33,6 +33,7 @@ export const AccountChangeContainer = () => {
     return <div>Loading ...</div>;
   }
 
+  //TODO: require PI or supervisor access after updating auth policies
   const changeAccounts = async () => {
     const response = await fetch(`/Request/Approve/${projectId}`, {
       method: "POST",
--- a/Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
@@ -38,13 +38,27 @@ export const AccountChangeContainer = () => {
     // if accounts look good, post them to new .net endpoint in Api/RequestController.cs
     // that endpoint will look something like the accounts portion of ApproveAsync method
     // on success, redirect back to project detail page
+    console.log(projectId, accounts);
+    const response = await fetch(`/Request/ApproveChange/${projectId}`, {
+      method: "POST",
+      headers: {
+        Accept: "application/json",
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ Accounts: accounts }),
+    });
+
+    // console.log(response)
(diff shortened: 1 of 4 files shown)


jSylvestre (changes requested):
One of mine was failing when it got to field info.
This threw an exception. You should notify the user when there is a problem.

jSylvestre:
Also, the approve quote page is broken in this PR (There is no longer an request/approve end point)

image

wongband:
@jSylvestre what do you mean by  (There is no longer an request/approve end point). Does the button not work anymore?

jSylvestre:
> @jSylvestre what do you mean by (There is no longer an request/approve end point). Does the button not work anymore?

@wongband Yes, this is pointing to request/approve . I don't know when it was changed.

Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
```diff
@@ -0,0 +1,85 @@
...
+      headers: {
+        Accept: "application/json",
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ Accounts: accounts }),
+    });
+
+    if (response.ok) {
+      const data = await response.json();
+      window.location.pathname = `/Project/Details/${data.id}`;
```
srkirkland, line 49: @wongband  we shouldn't re-route this way since it forces a full-page refresh.  Would be better to `useHistory` and push the new state, or use the <Redirect> component of react-router https://reactrouter.com/web/api/Hooks/usehistory.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Requests/AccountChangeContainer.tsx
srkirkland: If you want to do this in a separate PR we can.  There might be other places in the app doing this as well, and if so we should change them too.  I wrote a few like this from before those pages were react, but once all the pages became react we should then route doing client-side routing only.

wongband: Okay got it. I will change that

wongband: @srkirkland I fixed the routing in that component. I found a couple others as well. Do you want me to fix those in a separate PR?

srkirkland: @wongband perfect I think this is ready to merge and then yes, please fix the others in a different PR

srkirkland (approved):

PR #235, opened 2021-06-15
Auth for PI and simple usage

jSylvestre:
Part of Issue #217

### DIFF
--- a/Harvest.Core/Models/AccessCodes.cs
+++ b/Harvest.Core/Models/AccessCodes.cs
@@ -9,8 +9,8 @@ namespace Harvest.Core.Models
         public const string SupervisorAccess = "SupervisorAccess";
         public const string WorkerAccess = "WorkerAccess";
         /// <summary>
-        /// When using this Authentication, the project id must be in either the projectId or id parameters. It will look for the projectId first.
+        /// When using this Authentication, the projectId must be in the parameters.
         /// </summary>
-        public const string PrincipalInvestigatorProjectMustBeIdOrProjectId = "PrincipalInvestigator";
+        public const string PrincipalInvestigator = "PrincipalInvestigator";
     }
 }
--- a/Harvest.Core/Models/AccessCodes.cs
+++ b/Harvest.Core/Models/AccessCodes.cs
@@ -8,5 +8,6 @@ namespace Harvest.Core.Models
         public const string FieldManagerAccess = "FieldManagerAccess";
         public const string SupervisorAccess = "SupervisorAccess";
         public const string WorkerAccess = "WorkerAccess";
+        public const string PrincipalInvestigator = "PrincipalInvestigator";
     }
 }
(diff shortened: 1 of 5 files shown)


jSylvestre:
@wongband FYI, this PR does basic auth for looking at the PI of of project

Harvest.Web/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -30,12 +31,35 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
                 context.Succeed(requirement);
+                return;
             }
+
+            if (requirement.RoleStrings.Contains(Role.Codes.PI))
+            {
+                try
+                {
+                    //Assumes id will always be project id if PI string is used
+                    var projectId = Int32.Parse( _httpContext.HttpContext?.Request.RouteValues["id"]?.ToString());
```
sprucely, line 48: This seems like a convention that could bite us in the future. A developer placing `[Authorize(Policy = AccessCodes.PrincipalInvestigator)]` would need to remember that "id" must represent a projectId. I'm thinking like a route that is only indirectly related to a project, such as an id representing an invoice.

sprucely (approved):
Didn't test, but code looks okay.

Harvest.Web/Handlers/VerifyRoleAccessHandler.cs
jSylvestre: @sprucely Agreed. Only other thought is to always use projectId...

Harvest.Web/Handlers/VerifyRoleAccessHandler.cs
```diff
@@ -30,12 +31,44 @@ protected override async Task HandleRequirementAsync(AuthorizationHandlerContext
...
+                return;
             }
+
+            if (requirement.RoleStrings.Contains(Role.Codes.PI))
+            {
+                try
+                {
+                    int projectId = 0;
+                    //Try to get projectId Parameter:
+                    var projectString = _httpContext.HttpContext?.Request.RouteValues["id"]?.ToString();
```
srkirkland, line 49: I think you mean projectId here instead of Id

Harvest.Core/Models/AccessCodes.cs
```diff
@@ -8,5 +8,9 @@ public static class AccessCodes
         public const string FieldManagerAccess = "FieldManagerAccess";
         public const string SupervisorAccess = "SupervisorAccess";
         public const string WorkerAccess = "WorkerAccess";
+        /// <summary>
+        /// When using this Authentication, the project id must be in either the projectId or id parameters. It will look for the projectId first.
+        /// </summary>
+        public const string PrincipalInvestigatorProjectMustBeIdOrProjectId = "PrincipalInvestigator";
```
srkirkland, line 14: If we go with this method I don't think we need the crazy name

srkirkland (changes requested):
Usually we would do this as a security service instead of an attribute.  But I at least like the idea here, since it cleans up the code.  Let's talk about it as a group soon before merging.

Issue #212, opened 2021-06-03
Attach files to project request and project itself after creation


srkirkland:
Can be done by PI or sup or FM.

Should show on project details header.

- [x] Add files to request creation page
- [x] Add files from project details page
- [x] Allow removing unwanted files? If so, it should delete blob files too.
- [x] Show files on project details page

jSylvestre:
https://docs.microsoft.com/en-us/dotnet/api/overview/azure/storage.blobs-readme?view=azure-dotnet  ???
The processes Payments uses have been deprecated

jSylvestre:
SAS? 
https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview

srkirkland:
looks like we are done with this one

PR #240, opened 2021-06-17
Service for creating cloud file storage endpoints

srkirkland:
Simple service which generates upload/download links for azure files, with 1hr time limit.  Will be used in the future by project pages to storage/read files.

part of #212

### DIFF
--- a/Harvest.Web/Startup.cs
+++ b/Harvest.Web/Startup.cs
@@ -136,7 +136,9 @@ namespace Harvest.Web
             services.Configure<FinancialLookupSettings>(Configuration.GetSection("FinancialLookup"));
             services.Configure<SlothSettings>(Configuration.GetSection("Sloth"));
             services.Configure<SparkpostSettings>(Configuration.GetSection("SparkPost"));
+            services.Configure<StorageSettings>(Configuration.GetSection("Storage"));
 
+            services.AddSingleton<IFileService, FileService>();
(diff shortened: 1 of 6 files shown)


Harvest.Web/Startup.cs
```diff
@@ -136,7 +136,9 @@ public void ConfigureServices(IServiceCollection services)
             services.Configure<FinancialLookupSettings>(Configuration.GetSection("FinancialLookup"));
             services.Configure<SlothSettings>(Configuration.GetSection("Sloth"));
             services.Configure<SparkpostSettings>(Configuration.GetSection("SparkPost"));
+            services.Configure<StorageSettings>(Configuration.GetSection("Storage"));
 
+            services.AddSingleton<IFileService, FileService>();
```
sprucely, line 141: Why does this need to be a singleton?

jSylvestre (approved):
Seems a lot cleaner.
Will that create the container if it doesn't exist on upload?

srkirkland:
> Seems a lot cleaner.
> Will that create the container if it doesn't exist on upload?

No.  container needs to exist and have permissions to allow SAS tokens (which is on by default)

Harvest.Web/Startup.cs
srkirkland: Doesn't need to be but it has no state so I figured why bother recreating it on every request.  Though perhaps that isn't a good enough reason and since the others are all scoped I could do the same

Issue #223, opened 2021-06-04
Task/Ticketing system


srkirkland:
Ability for PIs (and their staff?) to be able to create work tickets associated with a project.  Simple requirements text, date due, completed flag (default to false).  Perhaps we allow some small back and forth with task comments similar to service now.  At some point the task is marked as completed.

Open tasks should show on the project details page, there should be a list of all tasks, and it would be a bonus if closed tasks were linked to the next invoice generated after they were closed.

From Brad:
> We talked about the “ticketing” system Israel mentioned.  Just FYI a reason I thought it would be useful to keep a running list of those with the project is for the record keeping reasons (billing and completed tasks) that Israel mentioned but also to append them in some way to the monthly billing statement would help the PIs and research staff have a better idea of what and how often they are requesting support (If I see that I requested labor or skilled staff help 10 times during the month, I should be less surprised by the bill for labor).

jSylvestre:
Also need to generate emails when things happen (Created, taskMessage added, completed)

jSylvestre:
This is pretty much done except for the save without notify.
We can revisit if there are issues.

PR #242, opened 2021-06-17
Add a Task table

jSylvestre:
Issue #223 
Just the tables

### DIFF
--- a/Harvest.Core/Domain/Ticket.cs
+++ b/Harvest.Core/Domain/Ticket.cs
@@ -19,17 +19,19 @@ namespace Harvest.Core.Domain
 
         [Required]
         public int CreatedById { get; set; }
+        public User CreatedBy { get; set; }
         public DateTime CreatedOn { get; set; }
 
-        public int UpdatedById { get; set; }
-        public DateTime UpdatedOn { get; set; }
+        public int? UpdatedById { get; set; }
+        public User UpdatedBy { get; set; }
+        public DateTime? UpdatedOn { get; set; }
 
         [StringLength(50)]
         [Display(Name = "Task Name")]
         public string Name { get; set; }
         [Required]
         public string Requirements { get; set; }
-        public DateTime DueDate { get; set; }
+        public DateTime? DueDate { get; set; }
 
         public string WorkNotes { get; set; } //For the person doing the Task
         [StringLength(25)]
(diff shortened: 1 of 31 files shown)


Harvest.Core/Domain/Ticket.cs
```diff
@@ -0,0 +1,70 @@
...
+        public int ProjectId { get; set; }
+        public Project Project { get; set; }
+        public int? InvoiceId { get; set; }
+        public Invoice Invoice { get; set; }
+
+        [Required]
+        public int CreatedById { get; set; }
+        public DateTime CreatedOn { get; set; }
+
+        public int UpdatedById { get; set; }
```
srkirkland, line 24: You'll need to make this int? For it to be optional

Harvest.Core/Domain/Ticket.cs
```diff
@@ -0,0 +1,70 @@
...
+        public DateTime UpdatedOn { get; set; }
```
srkirkland, line 25: Same here. Datetime needs to be nullable to be optional

Harvest.Core/Domain/Ticket.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        [StringLength(50)]
+        [Display(Name = "Task Name")]
+        public string Name { get; set; }
+        [Required]
+        public string Requirements { get; set; }
+        public DateTime DueDate { get; set; }
```
srkirkland, line 32: Maybe this should be optional too? Things might not always have a due date

Harvest.Core/Domain/Ticket.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public string WorkNotes { get; set; } //For the person doing the Task
```
srkirkland, line 34: Probably want to make this a Max length string

srkirkland (changes requested):

Harvest.Core/Domain/Ticket.cs
jSylvestre: It defaults to max. Doesn't seem to be a simple way of explicitly doing it.

srkirkland: i would think you could just do [MaxLength(int.Max)]` or in the prop.  not sure if there's a better way

srkirkland (approved):

PR #244, opened 2021-06-17
Replaced routing with useHistory

wongband:


sprucely (commented):
@srkirkland @jSylvestre For forms that redirect after a successful submit/save operation, should we be using `history.replace()` instead of `history.push()` so that the back button does not bring us back to a form that is no longer valid?

srkirkland:
> @srkirkland @jSylvestre For forms that redirect after a successful submit/save operation, should we be using `history.replace()` instead of `history.push()` so that the back button does not bring us back to a form that is no longer valid?

Probably a good idea, though in this case nothing would be invalid about using the back button since you can just resubmit

Harvest.Web/ClientApp/src/Requests/ApprovalContainer.tsx
```diff
@@ -47,7 +48,7 @@ export const ApprovalContainer = () => {
     });
 
     if (response.ok) {
-      window.location.pathname = `/Project/Details/${projectId}`;
+      history.push(`/Project/Details/${projectId}`);
```
srkirkland, line 51: @wongband @sprucely had a good point here -- use history.replace instead of .push since we don't want the person to hit the back button to come back to this page (though if they were to do that we should be able to handle it properly via the onload in the future)

srkirkland (approved):

Harvest.Web/ClientApp/src/Requests/ApprovalContainer.tsx
wongband: Okay I can change that @srkirkland should I also change it in the ExpenseContainer and QuoteContainer since they also save/submit in a form?

srkirkland: I would say we don't have to bother since there is no real issue with submitting multiple requests or expenses and id like to err on the side of not replacing browsing history.

wongband: okay sounds good

PR #245, opened 2021-06-18
Client-side file upload

srkirkland:
Client side of file upload v1 completed.  Doesn't save to db but this seemed like a good stopping point.  Uses the normal html file input to get files, uploads them to blob storage, and then tracks and validates they have been uploaded before allowing the user to click submit.  The file info is added to the project type so saving/retrieving should be simple from here.

part of #212


jSylvestre (approved):
Was able to upload 673e27a3-2d5d-47ca-940e-38b5e20594aa and manually get the file from azure

PR #247, opened 2021-06-21
set content type and disposition of new blob

srkirkland:


jSylvestre:
@srkirkland I see the content type, but I don't see anything that indicates a filename. And If I try to download directly from azure, it just uses the identifier as the name:

image

srkirkland:
@jSylvestre My guess is that when you download via the sas token (or if we made the endpoint public) then it would respect the content disposition header and send the right file name. The clients will be downloading via sas endpoint so as long as that works it should be good.

jSylvestre (approved):

PR #248, opened 2021-06-22
Cyd/account details

cydoval:
Screen Shot 2021-06-22 at 12 00 18 PM

There is code that probably needs to be moved around, specifically the total percent row/col AccountsInput.tsx, it should only show when an account has been added. I couldn't figure out how to get it to do that without it being all red


cydoval:
And the static text needs to be hooked up/changed to be dynamic

jSylvestre:
Don't we want the account name?

srkirkland (approved):

jSylvestre (approved):

PR #249, opened 2021-06-22
Jcs/work items controller

jSylvestre:
Doesn't do attachments (yet).
Probably lots to fix, but it is working


Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
```diff
@@ -0,0 +1,130 @@
...
+            headers: {
+                Accept: "application/json",
+                "Content-Type": "application/json",
+            },
+            body: JSON.stringify(createTicket),
+        });
+
+        if (response.ok) {
+            const data = await response.json();
+            window.location.pathname = `/Project/Details/${data.id}`;
```
srkirkland, line 55: use history here to push state

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
```diff
@@ -0,0 +1,130 @@
...
+                setProject(proj);
+                setCreateTicket({...createTicket, projectId:proj.id});
+            }
+        };
+
+        cb();
+    }, [projectId]);
+
+    if (project === undefined) {
+        return <div>Project Not Found!</div>;
```
srkirkland, line 37: project will be undefined if the fetch hasn't come back yet too, so it's not always because project not found.

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
```diff
@@ -0,0 +1,130 @@
...
+                        onChange={(e) =>
+                            setCreateTicket({ ...createTicket, name: e.target.value })
+                        }
+                        placeholder="Enter a short description for this request"
+                    />
+                </FormGroup>
+                
+                <FormGroup>
+                    <Label>What are the details of your ticket request?</Label>
+                    <Input
```
srkirkland, line 91: this would probably be better as a `<textarea>` than an `input`

Harvest.Web/ClientApp/src/types.ts
```diff
@@ -215,6 +215,13 @@ export interface ProjectAccount {
   percentage: number;
 }
 
+export interface CreateTicket {
```
srkirkland, line 218: Just `Ticket`

Harvest.Web/ClientApp/src/types.ts
```diff
@@ -215,6 +215,13 @@ export interface ProjectAccount {
...
+    projectId: number;
+    name: string;
+    requirements: string;
+    dueDate: Date;
```
srkirkland, line 222: Is dueDate optional?

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -0,0 +1,66 @@
...
+        public ActionResult Create()
+        {
+            return View("React");
+        }
+
+        [HttpPost]
+        [Authorize(Policy = AccessCodes.PrincipalInvestigator)]
+        public async Task<ActionResult> Create(int projectId, [FromBody] Ticket ticket)
+        {
+            if (projectId != ticket.ProjectId)
```
srkirkland, line 38: this is ok but if it was me I wouldn't bother checking and I'd just set ticket.projectId to the projectId.  And I wouldn't bother setting it in client code.

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                return BadRequest("Wrong Project Id");
+            }
+
+            var project = await _dbContext.Projects.SingleAsync(a => a.Id == projectId);
```
srkirkland, line 43: why bother getting project?

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
jSylvestre: @srkirkland I believe that is how it works. it has a type of textarea

Harvest.Web/Controllers/Api/TicketController.cs
jSylvestre: @srkirkland I'm returning to the project details

srkirkland: wouldn't it make more sense to return the ticket details since it's to create a new ticket?

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
srkirkland: Oh i see, because it's not a real `<input>` but one of the reactstrap ones

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
```diff
@@ -0,0 +1,125 @@
...
+import DatePicker from "react-date-picker";
+import { Button, FormGroup, Input, Label } from "reactstrap";
+
+interface RouteParams {
+  projectId?: string;
+}
+
+export const TicketContainer = () => {
+  const { projectId } = useParams<RouteParams>();
+  const [project, setProject] = useState<Project | undefined>();
```
srkirkland, line 14: @jSylvestre curious as to why you did this?  If you do `useState<Project>()` since you leave the initializer empty it knows the `project` is possibly undefined and you don't need to add the `| undefined` part

srkirkland (approved):
take a look at your comments, especially in the react file.  I think most were just copy pasted and make no sense given the context

Harvest.Web/Controllers/Api/TicketController.cs
jSylvestre: I was copying code. I don't have any other ticket pages at the moment.

Harvest.Web/ClientApp/src/Tickets/TicketContainer.tsx
jSylvestre: Was just coping code. Fixed.

PR #250, opened 2021-06-22
Client-side user permission checks

srkirkland:
Goal is to be able to show/hide links in React depending on user roles, as well as restrict a page entirely if they don't have the proper role.

Whenever the Layout page is loaded (should be just once because of React routing, but can be forced via a refresh), the user's roles are pulled from the DB and shoved into the page.  This is read into a new AppContext, which might eventually include some additional user or config info.

From there, I created a ConditionalRoute component which just takes in a list of roles that can see the given route, and it'll handle restricting the route as needed.


Harvest.Web/Views/Shared/_Layout.cshtml
```diff
@@ -112,6 +112,12 @@
             All rights reserved.</p>
     </footer>
 
+
+    <script>
+        var Harvest = { user: {} };
+        Harvest.user.roles = @Html.Raw(await Context.GetUserRoles());
+    </script>
+
```
sprucely, line 120: Do roles really need to be pushed in this way? Why not just have a UserRolesPromise initialize on the app page and pass it into AppContextProvider?

srkirkland: Could be done either way.  This just seemed easier to me because you have one less http request and one less promise loading state.

sprucely: Hmm. This does eliminate an extra round trip.

sprucely (approved):

Harvest.Web/ClientApp/src/types.ts
```diff
@@ -174,7 +180,7 @@ export class WorkItemImpl implements WorkItem {
   type;
   rate;
   rateId = 0;
-  unit = 'hr';
+  unit = "hr";
```
jSylvestre, line 183: @srkirkland Do we have or need a prettier file?

Harvest.Web/Views/Shared/_Layout.cshtml
jSylvestre: Dumb question. Could they just edit this? Or we will only use this to show and hide stuff and it will get reloaded each time in that case.

jSylvestre (approved):

Harvest.Web/ClientApp/src/types.ts
srkirkland: I think it was prettier that formatted this

Harvest.Web/Views/Shared/_Layout.cshtml
srkirkland: @jSylvestre they can't really edit that, but they can easily edit the javascript that reads it.  But yes, this is only to show/hide, we won't be relying on any of this for actual security.

Issue #219, opened 2021-06-03
Page where unbilled expenses are editable 


srkirkland:

srkirkland:
instead we are going to do just view and delete

PR #254, opened 2021-06-23
start of un-billed page with activity rehydrator 

srkirkland:
I'm having second thoughts on making the unbilled expenses all editable.  I think it'll be quite doable technologically but it'll loose the history of who created the expense and we don't have editor tracking.  Perhaps it would be better to just display the expenses and then allow deleting and maybe single-item editing, since hopefully you shouldn't be regularly bulk-editing all of your already entered expenses.

This PR so far contains some nice code for returning unbilled expenses data and transforming them into activities, so I don't want to loose it.  But let's all talk about how we want to handle this before I got any futher.  Part of #219.


srkirkland:
decided instead to have un-billed table with only delete allowed

PR #256, opened 2021-06-23
Created PDF for Quote

### LINKED ISSUE #239: Create PDF of quote
Goal is to be able to generate PDFs, ideally on the client and then downloadable directly.  First use-case is for the request approval page, ex https://localhost:5001/request/approve/25.  When they click the "Download PDF" link it should make a PDF that looks more or less like the Quote on the page (QuoteDisplay component).

Don't have a favorite PDF creation library yet, but https://react-pdf.org/ looks pretty good.

wongband:
resolve #239 

Installed the following packages:
- react-pdf
- react-pdf-table


sprucely (approved):
Worked for me. Not the prettiest, but maybe we can let Cal deal with styles later. Pretty cool!

srkirkland (approved):
good start -- we'll probably change some names later and maybe we can punch up the formatting but i think it's already a really nice feature

PR #261, opened 2021-06-24
progress bars on project details and list pages are now hooked up

srkirkland:


jSylvestre (approved):
Works for me by manually changing my data.
Do we want a color change when we get close to the quote total?

jSylvestre:
Did you want to change this in the PR as well?

image

srkirkland:
> Works for me by manually changing my data.
> 
> Do we want a color change when we get close to the quote total?

I guess we can ask @cydoval about that one but this will be a start

PR #262, opened 2021-06-24
cleaned up random console warnings & useEffect minor issues

srkirkland:
I left the anchor tag warnings since we'll probably want to get those soon in a real PR, but otherwise I cleaned up a bunch of minor console warnings.


jSylvestre (approved):
Didn't test, but looks ok to me.

PR #270, opened 2021-06-25
Added date validation check

### LINKED ISSUE #265: when creating request, start date must be before end date
probably a simple yup validation but we should not be able to submit the form without that condition

wongband:
resolve #265


jSylvestre (changes requested):
Steps to reproduce:
1) Create new request and click Create Field Request without entering anything
2) Enter all fields except start and end, then enter start and end where end is before start.
3) Click create button again:

image

jSylvestre:
I don't think my change request was introduced by this PR... But probably good to fix anyway.

jSylvestre (approved):

PR #272, opened 2021-06-29
show just recent invoices but provide a link to full list

srkirkland:
currently full invoice list isn't linked to a aspnet route, so that should be done once routes are updated to allow for persistent links.


Harvest.Web/ClientApp/src/Invoices/InvoiceListContainer.tsx
```diff
@@ -1,28 +1,36 @@
...
 
       if (response.ok) {
         setInvoices(await response.json());
       }
     };
 
+    if (projectId) {
+      cb();
+    }
     cb();
```
sprucely, line 28: if projectId, callback, else callback?

srkirkland:
Oooooops
________________________________

PR #273, opened 2021-06-29
TicketTable

jSylvestre:


Harvest.Web/ClientApp/src/Tickets/TicketTable.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const TicketTable = (props: Props) => {
+  const ticketData = useMemo(() => props.tickets, [props.tickets]);
+  const columns: Column<Ticket>[] = useMemo(
+      () => [
+          {
+              Cell: (data: Cell<Ticket>) => (
+                  <div>
+                      <a href={`/ticket/details??projectId=${data.row.original.projectId}&id=${data.row.original.id}`}>
```
srkirkland, line 18: need to use the `<Link>` not the `<a>` if it's a react route

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -23,6 +23,21 @@ public TicketController(AppDbContext dbContext, IUserService userService)
             this._userService = userService;
         }
 
+        // create a new ticket via react
+        [HttpGet]
+        [Authorize(Policy = AccessCodes.PrincipalInvestigator)]
+        public async Task<ActionResult> ListTickets(int projectId, bool topOnly = true)
```
srkirkland, line 29: you don't need to put `Tickets` in the name if it's already the ticket controller.  `/ticket/list/` i think is enough info on what is happening

Harvest.Web/ClientApp/src/Tickets/TicketListContainer.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useEffect, useState } from "react";
+import { useHistory, useParams } from "react-router-dom";
+
+import { Ticket } from "../types";
+import { TicketTable } from "./TicketTable";
+
+interface Props {
+  projectId: any;
+}
+interface RouteParams {
```
srkirkland, line 10: from what i can tell this is never being used/populated.  You seem to be using this from another component and passing props, and if that's the case you should be using that instead of half-using the props.projectId and half using projectId.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Tickets/TicketTable.tsx
jSylvestre: @srkirkland I based this off the invoice table. Want me to update that one too?

srkirkland: @jSylvestre maybe but I have a PR open which changes the invoices table a lot so be careful, or do it later.

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -45,6 +46,7 @@ function App() {
       <ConditionalRoute roles={['FieldManager', 'Supervisor']} path="/quote/create/:projectId" component={QuoteContainer} />
       <Route exact path="/project" component={ProjectListContainer} />
       <Route path="/ticket/create/:projectId" component={TicketContainer} />
+      <Route path="/ticket/allTickets/:projectId" component={AllTickets} />
```
srkirkland, line 49: just call this `/ticket/all/123`, or /ticket/list/123.  you don't need the duplication on `/ticket/allTickets`.

srkirkland: Also this is minor and i'm not even sure why we do it, but top level things we call XYZContainer.  So to keep with our naming scheme we would probably call this `TicketListContainer` or perhaps just `TicketsContainer`

Harvest.Web/ClientApp/src/Tickets/AllTickets.tsx
```diff
@@ -0,0 +1,69 @@
...
+        const proj: Project = await response.json();
+        setProject(proj);
+      }
+    };
+
+    cb();
+  }, [projectId]);
+
+  useEffect(() => {
+    // get rates so we can load up all expense types and info
```
srkirkland, line 31: dump, obsolete comment from copy/paste

Harvest.Web/ClientApp/src/Tickets/AllTickets.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+      if (response.ok) {
+        setTickets(await response.json());
+      }
+    };
+
+    cb();
+  });
+
+  if (project === undefined || tickets === undefined) {
```
srkirkland, line 45: Don't think tickets can ever be undefined since you give it a default value

Harvest.Web/ClientApp/src/Tickets/AllTickets.tsx
```diff
@@ -0,0 +1,69 @@
...
+      if (response.ok) {
+        const proj: Project = await response.json();
+        setProject(proj);
+      }
+    };
+
+    cb();
+  }, [projectId]);
+
+  useEffect(() => {
```
srkirkland, line 30: you should make this rely on the projectId just like the other useEffect.  If the projectId changes, you definitely need to fetch new tickets, you can't just do it once.

Harvest.Web/ClientApp/src/Tickets/TicketListContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { Ticket } from "../types";
+import { TicketTable } from "./TicketTable";
+
+interface Props {
+  projectId: any;
+}
+
+
+export const TicketListContainer = (props: Props) => {
```
srkirkland, line 13: This isn't a container anymore now that it is being passed props and it's top level.  Better to just call it `TicketList` or similar to make room from your new container.

Harvest.Web/ClientApp/src/Tickets/TicketTable.tsx
```diff
@@ -0,0 +1,68 @@
...
+        maxWidth: 150,
+      },
+      {
+        Header: "Status",
+        accessor: (row) => row.status,
+      },
+      {
+        Header: "Created",
+        accessor: (row) =>
+          row.createdOn ? new Date(row.createdOn).toLocaleDateString() : "",
```
srkirkland, line 37: is created really optional?

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -23,6 +23,26 @@ public TicketController(AppDbContext dbContext, IUserService userService)
             this._userService = userService;
         }
 
+        // create a new ticket via react
+        [HttpGet]
+        [Authorize(Policy = AccessCodes.PrincipalInvestigator)]
+        public async Task<ActionResult> List(int projectId, bool topOnly = true)
```
srkirkland, line 29: I'd rather you optionally pass the number you want instead of `topOnly`.  it can be an optional field and if it's null get them all.  This was Cal can decide if he wants 3 or 5 or more and we only have to change one thing.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Tickets/TicketTable.tsx
jSylvestre: No, but i wasn't sure when I added it to types if it wasn't optional if it would cause problems with the create ticket where this gets set in the controller.

Harvest.Web/ClientApp/src/Tickets/AllTickets.tsx
jSylvestre: @srkirkland Do you mean just put the `, [projectId]` back in this one like it is above?

srkirkland: yes

Harvest.Web/ClientApp/src/Tickets/RecentTicketsContainer.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useEffect, useState } from "react";
+import { Link } from "react-router-dom";
+
+import { Ticket } from "../types";
+import { TicketTable } from "./TicketTable";
+
+interface Props {
+    projectId: any;
```
srkirkland, line 8: shouldn't be any here.  i think it's a string right?

Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <h2>Ticket Attachments</h2>
+        <p>No Attachments</p>
+      </div>
+    );
+  }
+
+  return (
+    <div>
+      <h2>Ticket Attachments</h2>
+      <p>{ticketAttachments[0].fileName}</p>
```
srkirkland, line 25: i'm assuming this part is a work in progress?

Harvest.Web/ClientApp/src/Tickets/TicketsContainer.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+      if (response.ok) {
+        setTickets(await response.json());
+      }
+    };
+
+    cb();
+  }, [projectId]);
+
+  if (project === undefined) {
```
srkirkland, line 41: not sure the best way to handle this, but you'll probably want to check that both tickets and projects have loaded.  As it is now, once the project loads then loading will go away and it could show 0 tickets even though we are still waiting for the ticket API call to come back.  Perhaps you'll want to not initialize to `[]` and check for undefined on the tickets array too?

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using Harvest.Core.Data;
 using Harvest.Core.Domain;
+using Harvest.Core.Migrations.SqlServer;
```
srkirkland, line 7: why?

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -72,5 +111,57 @@ public async Task<ActionResult> Create(int projectId, [FromBody] Ticket ticket)
...
+        {
+            return View("React");
+        }
+        [HttpGet]
+        [Route("[controller]/[action]/{projectId}/{ticketId}")]
+        public async Task<ActionResult> Get(int projectId, int ticketId)
+        {
+            var ticket = await _dbContext.Tickets
+                .Where(a => a.Id == ticketId && a.ProjectId == projectId)
+                .Select(a => new
```
srkirkland, line 125: 👍

srkirkland (approved):
i think you are getting the hang of this

PR #274, opened 2021-06-30
Created pdf for Invoices and made PDF components more flexible

### LINKED ISSUE #266: ability to download an invoice as a PDF
should be very similar to the quote PDF and could possibly use many of the same components.  Download button should be somewhere on the view invoice page, ex: invoice/details/9

wongband:
resolve #266


jSylvestre (approved):

srkirkland (approved):
awesome looks great

PR #275, opened 2021-06-30
ShowFor

jSylvestre:


srkirkland (commented):
Yea, I think any is a fine starter for the children. It can easily be changed later. And for your show var you can just have it be like a string array of roles and your component can do a useEffect or useMemo to calculate the value of shouldShow based off of the users roles within the context api

srkirkland (approved):
This looks pretty good to me but i think it's worth mentioning that every time this component re-renders (which could happen a lot if, for instance, if it's part of some form where inputs are changing) the user roles search is going to be recalculated, even though it never changes.  This is not really worth fixing here since a small array search is very fast, but i wanted to mention it because there will be other cases where it is more important.  A few ways to fix this but the quick one would be for your `anyMatchingRoles` variable to be a `useMemo()` which only is re-calculated when the values in `props.roles` changes.

PR #277, opened 2021-07-02
Added FileUpload and displayed attachments to ProjectDetailContainer

### LINKED ISSUE #257: hook up attachments on project page.
on the project details page, ex /project/details/28, hook up the attachments section to show the actual attachments associated with the project.  Also add in the fileUpload.tsx file so they can upload new files just like how it works on the new request page.  If a new file is upload it should show up in the list.

wongband:
resolve #257

### DIFF
--- a/Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
@@ -9,7 +9,7 @@ import { FileUpload } from "../Shared/FileUpload";
 import { ProjectHeader } from "../Shared/ProjectHeader";
 import { RecentInvoicesContainer } from "../Invoices/RecentInvoicesContainer";
 import { ProjectUnbilledButton } from "./ProjectUnbilledButton";
-import { Project } from "../types";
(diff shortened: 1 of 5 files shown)


jSylvestre (changes requested):
I'm not seeing the attachments I created in the original request.

jSylvestre (changes requested):
Now you need to save the newly uploaded files on the project details page to the database.
Simplest way would be to add a new submit button to that page and an endpoint to add the attachments.
The other thing that might be nicer is to detect when the file(s) has been uploaded, then automatically call a new endpoint to add it to the DB.

wongband:
@jSylvestre I updated it so that it updates automatically whenever a new file is uploaded

jSylvestre (approved):
Looks ok to me, but wait for @srkirkland before merging.

And just a FYI, should the files that have already been uploaded show up here?
image

srkirkland (approved):
Code looks good.   Your formatting seems to be kinda strange.  Are you using `prettier` and is it configured to use the version of prettier in the clientApp folder?  Is there some other formatting extension you are using?

wongband:
@srkirkland I think I am using the prettier config. I will double check

PR #282, opened 2021-07-07
What am I doing wrong?

jSylvestre:


jSylvestre:
I wanted to make the edit worknotes a component and just pass the ticket in the parameters.
It is undefined

PR #284, opened 2021-07-08
Added delete column to un-billed expenses column

### LINKED ISSUE #259: add the ability to delete a row from unbilled expenses
viewing any unbilled expenses, ex: /expense/unbilled/25, add a column to allow deleting the row.  there should be a confirmation required, but it doesn't have to be a fancy one.  Delete the row on the db on confirm and then it should remove itself from the table too.

needs to be done after #258 is merged

wongband:
resolve #259


srkirkland (approved):

Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
```diff
@@ -50,9 +74,34 @@ export const UnbilledExpensesContainer = () => {
...
+            <Button
+              color="success"
+              onClick={() =>
+                selectedExpense !== null
+                  ? confirmModal(selectedExpense.id)
+                  : null
+              }
+            >
+              Confirm
+            </Button>{" "}
```
jSylvestre, line 93: What is the {" "} for?

jSylvestre (approved):

Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
wongband: It creates space between the two buttons

jSylvestre: Hmm, I removed it and didn't notice any difference.
Probably better to use a class for this kind of formatting. @cydoval ?

srkirkland: Empty string is actually pretty common way to do things. But yea there might be something better. But you do see it this way a lot, especially when prettier reformats things.

PR #285, opened 2021-07-08
Removed spinner for uploaded attachments

wongband:


jSylvestre:
@wongband These show up when you open a project that already has attachments.
Can you only show attachments here when they are newly uploaded? Or maybe only as they are uploading?
image

jSylvestre (approved):

wongband:
@jSylvestre I have an idea on how to implement that. I'll try it out right now

srkirkland (approved):

srkirkland (approved):

PR #288, opened 2021-07-12
default to projectId for most routes

### LINKED ISSUE #238: Update startup route to default to {projectId} instead of ID
then update all methods and handle the new variable.  anything not using projectId will need to update as well.

Then update policy

srkirkland:
other than a few controllers, the default route is now {controller}/{action}/{projectId} and you must take in `int projectId` in those methods.  

It didn't take too many changes to do it, and it'll now make it unambiguous when checking permissions which route param is the project id.

closes #238


jSylvestre (approved):
Didn't test, but looks good so far.

Issue #220, opened 2021-06-03
Big changes to Quote page


srkirkland:
- [x] Add year dropdown to activity
- [x] Add adjustment %, should adjust all rates
- [x] Default adjustment based on 3% per year
- [x] Make all rates readonly
- [x] Duplicate activity button
- [x] Add readonly box to acreage fee area for num years

PR #289, opened 2021-07-12
Make rate not editable anywhere

srkirkland:
also added a prop so you can hide the annual adjustment header if it's not needed

part of #220


sprucely (approved):
Didn't test, but looks fine to me.

PR #290, opened 2021-07-12
Removed old files in ProjectDetailsContainer

wongband:


jSylvestre:
Haven't tested it yet, but does this work correctly if the select more than 1 attachment at a time?

srkirkland:
> Haven't tested it yet, but does this work correctly if the select more than 1 attachment at a time?

I have the same question given the `f[f.length-1]` portion

wongband:
Sorry about that I will test that out right now for multiple attachments

jSylvestre (changes requested):
Looks like you didn't save both attachments:

Selected 2 files:
image

Then, refresh the page:
image

jSylvestre:
@wongband FYI, I'm not saying this is the best way to do it, but this is how I added multiple attachments for tickets server side:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Web/Controllers/Api/TicketController.cs#L173

wongband:
@jSylvestre I will take a look at it and refactor

jSylvestre (changes requested):
After you have uploaded your files to request\files you need to clear out newFiles (I think)
Otherwise if you upload 2 files, then upload another file your model is passed 3 files instead of just the file that was newly selected.

jSylvestre (approved):
Works for me.

PR #292, opened 2021-07-12
Ability to do adjustments based on year

srkirkland:
If a request covers >1 year, we show a year selector at the top of each activity, and activities now contain year and adjustment (in percentages, multiples of 3%).  Change this selector adjusts all sub-work items by the given percent.

WorkItems themselves don't change, which is nice because it means we can continue to re-use this component for expense entry.

does the rest of #220 except for duplication and acreage rate


sprucely (approved):
Works for me and seems straight forward.

PR #293, opened 2021-07-13
Acreage Rate must include years

srkirkland:
Acres * Rate * years of project (min 1).  Added years to quote itself, so we don't have to keep calculating it from the project.  Seems to persist properly across saves.

part of #220


jSylvestre (approved):
Ok. But the approve quote needs year info too:

image

PR #296, opened 2021-07-13
update status with a classname

srkirkland:
also add some ids and accessors in the way react table likes


jSylvestre:
@srkirkland Since you are in here, did you want to show the project name in there when a quote is created?

srkirkland:
maybe when accepted?  definitely not created but i could see accepted being good.  I'm going to just make an issue of this as something to ask in a future meeting -- who gets to make the names and when, assuming we even need names

jSylvestre (approved):

PR #300, opened 2021-07-14
default PI to current user when creating new request

### LINKED ISSUE #241: Default PI to creator
When doing a field request, default the PI to the creator so they don't have to pick themselves

srkirkland:
Added to the layout serialized user context, so now in addition to roles we always know the current user details (could come in handy in other scenarios).  When creating a new request we just take the user details out of the context and make them the default PI.

closes #241

### DIFF
--- a/Harvest.Web/ClientApp/src/Test/mockData.ts
+++ b/Harvest.Web/ClientApp/src/Test/mockData.ts
@@ -1,7 +1,20 @@
 import { AppContextShape, ProjectWithQuote, Rate } from "../types";
 
+const fakeUser = {
+  id: 1,
+  firstName: "Bob",
+  lastName: "Dobalina",
+  email: "bdobalina@ucdavis.edu",
+  iam: "1000037182",
+  kerberos: "bdobalina",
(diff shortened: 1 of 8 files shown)


Harvest.Web/ClientApp/src/Test/mockData.ts
```diff
@@ -1,7 +1,20 @@
 import { AppContextShape, ProjectWithQuote, Rate } from "../types";
 
+const fakeUser = {
+  id: 1,
+  firstName: "Bob",
+  lastName: "Dobalina",
```
jSylvestre, line 6: Been a while since I've seen Dobalina

jSylvestre (changes requested):
Validation:

image

srkirkland:
@jSylvestre this first name thing is an interesting one -- in particular it's a CAS problem, since CAS should provide first name but it isn't.  But still, we should still handle this case.  I'll look into maybe returning an empty string or something if this happens.

jSylvestre (approved):
Sting would approve :)

PR #302, opened 2021-07-14
Jcs/ticket attachments

jSylvestre:


Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
```diff
@@ -1,28 +1,82 @@
...
+              Accept: "application/json",
+              "Content-Type": "application/json",
+          },
+          body: JSON.stringify({ Attachments: attachments }),
+        });
+
+        if (response.ok) {
+            const data = await response.json() as TicketAttachment[];
+
+            setTicket({ ...ticket, attachments: [...ticket.attachments, ...data] });
```
srkirkland, line 35: personally I would rather have you keep `setTicket` as `props.setTicket` just so it's clear in this case since it's competing with a local setTicketLoc which is so similar.  Also really we probably need to find a better way to track this new files thing, having a shadow local state var is not great.

srkirkland (approved):
I don't really like this but I know it's mostly a copy/paste of how we did things elsewhere, so I'll approve for now.  Then maybe we'll remember to come back around and do better about this whole local/new files thing

Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
jSylvestre: Yeah, agree. But I was having problems getting it to work.

PR #305, opened 2021-07-15
cleanup

jSylvestre:


Harvest.Web/ClientApp/src/Tickets/TicketWorkNotesEdit.tsx
```diff
@@ -26,7 +26,7 @@ export const TicketWorkNotesEdit = (props: Props) => {
       );
 
       if (response.ok) {
-          const data = await response.json();
+          await response.json();
```
srkirkland, line 29: What is the purpose of this?  It's just taking time to do nothing.

srkirkland (approved):

Harvest.Web/ClientApp/src/Tickets/TicketWorkNotesEdit.tsx
jSylvestre: I pop up an alert when it is saved. I can remove that if you want

jSylvestre: Oh, I see what you mean...

PR #306, opened 2021-07-15
Jcs/packages20210715

jSylvestre:
Still seeing 
[Deprecation] SharedArrayBuffer will require cross-origin isolation as of M92, around July 2021

with these changes...


srkirkland:
@jSylvestre I vote we just close without merging and then do a PR a little later that updates more related libraries.  Just updating this one isn't really doing anything and doesn't fix that warning either.

PR #307, opened 2021-07-15
Swe/access

### LINKED ISSUE #281: startup.cs access codes
Codes don't make sense or are wrong:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Web/Startup.cs#L96

sprucely:
closes #281

- Extend `IUserService` to allow checking roles/access in business logic.
- Extend `<ShowFor />` with optional `condition` property defaulting to `true`
- Hide most UI elements that are not valid for user's given access level. For some, disabling may look nicer. I added a `useRoleChecks` hook that provides functions `anyMatchingRoles`, `hasAccessForRoles` that can help, but nothing is using them yet.


Harvest.Core/Data/DbInitializer.cs
```diff
@@ -25,10 +25,10 @@ public async Task Initialize(bool recreateDb)
 
             //Make sure roles exist
             await CheckCreateRole(Role.Codes.System);
-            await CheckCreateRole(Role.Codes.Admin);
             await CheckCreateRole(Role.Codes.FieldManager);
             await CheckCreateRole(Role.Codes.Supervisor);
             await CheckCreateRole(Role.Codes.Worker);
+            await CheckCreateRole(Role.Codes.PI);
```
jSylvestre, line 31: I don't think this is an actual role. Shouldn't be in the DB.

jSylvestre: Actually ignore this. Sounds like we are going in this direction

srkirkland: no i agree here, I am not sure it should be in the DB or an actual role

Harvest.Core/Services/UserService.cs
```diff
@@ -61,6 +71,47 @@ public async Task<User> GetCurrentUser()
...
+        {
+            var roles = _roleResolver(accessCode).Concat(_roleResolver(AccessCodes.SystemAccess));
+            var userRoles = await GetCurrentRoles();
+            return userRoles.Any(r => roles.Contains(r));
+        }
+
         public const string IamIdClaimType = "ucdPersonIAMID";
     }
+
+    public static class UserServiceExtensions
```
srkirkland, line 94: Why are these extensions and not just part of the user service?  Extensions are nice when we don't control the class, like string or date extensions.  don't get the need for them here.

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -30,30 +30,33 @@ function App() {
     <AppContext.Provider value={Harvest}>
       <Route exact path="/" component={Home} />
       <Route exact path="/home/spa" component={Spa} />
-      <Route path="/request/create/:projectId?" component={RequestContainer} />
-      <Route path="/request/approve/:projectId" component={ApprovalContainer} />
-      <Route
+      <ConditionalRoute roles={["FieldManager"]} path="/request/create/:projectId?" component={RequestContainer} />
+      <ConditionalRoute roles={["FieldManager"]} path="/request/approve/:projectId" component={ApprovalContainer} />
```
srkirkland, line 34: The PI is probably the only one who should be approving requests.  Maybe a FM should be able to in emergency scenarios, but the PI is the main one.

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -30,30 +30,33 @@ function App() {
     <AppContext.Provider value={Harvest}>
       <Route exact path="/" component={Home} />
       <Route exact path="/home/spa" component={Spa} />
-      <Route path="/request/create/:projectId?" component={RequestContainer} />
-      <Route path="/request/approve/:projectId" component={ApprovalContainer} />
-      <Route
+      <ConditionalRoute roles={["FieldManager"]} path="/request/create/:projectId?" component={RequestContainer} />
```
srkirkland, line 33: why can't non-FMs create a project?  shouldn't everyone be able to?

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -30,30 +30,33 @@ function App() {
...
+      <ConditionalRoute roles={["FieldManager"]} path="/request/approve/:projectId" component={ApprovalContainer} />
+      <ConditionalRoute
+        roles={["FieldManager"]}
```
srkirkland, line 36: this should be primarily by the PI

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -30,30 +30,33 @@ function App() {
...
         path="/request/changeAccount/:projectId"
         component={AccountChangeContainer}
       />
       <Route
         path="/project/invoices/:projectId"
         component={InvoiceListContainer}
       />
-      <Route
+      <ConditionalRoute
+        roles={["FieldManager"]}
```
srkirkland, line 45: also primarily PI.  But probably supervisor too.

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -30,30 +30,33 @@ function App() {
...
         path="/invoice/details/:invoiceId"
         component={InvoiceDetailContainer}
       />
-      <ConditionalRoute roles={['FieldManager', 'Supervisor']} path="/quote/create/:projectId" component={QuoteContainer} />
+      <ConditionalRoute roles={["FieldManager"]} path="/quote/create/:projectId" component={QuoteContainer} />
```
srkirkland, line 49: supervisors need to be able to create a quote too.  Probably they shouldn't be able to submit them, but that's a separate question

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
```diff
@@ -61,31 +62,32 @@ export const ProjectDetailContainer = () => {
         <div className="card-content">
           <div className="row justify-content-between">
             <div className="col">
-              {project.status === "Requested" && (
+              <ShowFor roles={["FieldManager"]} condition={project.status === "Requested"}>
```
srkirkland, line 65: supervisor as well

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
```diff
@@ -61,31 +62,32 @@ export const ProjectDetailContainer = () => {
...
                   to={`/quote/create/${project.id}`}
                 >
-                  Edit Quote
+                    Edit Quote
                 </Link>
-              )}
+              </ShowFor>
 
-              {project.status === "PendingApproval" && (
+              <ShowFor roles={["FieldManager"]} condition={project.status === "PendingApproval"}>
```
srkirkland, line 74: PI

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
```diff
@@ -61,31 +62,32 @@ export const ProjectDetailContainer = () => {
...
-              <Link
-                className="btn btn-primary btn-small mr-4"
-                to={`/request/changeAccount/${project.id}`}
-              >
-                Change Accounts
-              </Link>
-              )}
+              </ShowFor>
+
+              <ShowFor roles={["FieldManager"]} condition={project.status === "Active"}>
```
srkirkland, line 83: PI

Harvest.Web/ClientApp/src/Shared/ShowFor.tsx
```diff
@@ -2,25 +2,26 @@ import React, { useContext, useMemo } from "react";
...
 interface Props {
   children: any;
   roles: RoleName[];
+  condition?: boolean | (() => boolean);
 }
 
 export const ShowFor = (props: Props) => {
   const { children, roles } = props;
   const userRoles = useContext(AppContext).user.roles;
+  const conditionSatisfied = isBoolean(props.condition) ? props.condition : isFunction(props.condition) ? props.condition() : true;
```
srkirkland, line 16: either write out or throw some parenthesis in there.  i don't like the visually ambiguous double-ternary operator.

Harvest.Web/Controllers/ProjectController.cs
```diff
@@ -31,19 +31,31 @@ public ActionResult Index()
...
+                .ToArrayAsync();
+            var hasNonPiRoles = userRoles.Any(r => r != Role.Codes.PI);
+            var isPi = userRoles.Contains(Role.Codes.PI);
+
             // TODO: only show projects where between start and end?
-            return Ok(await _dbContext.Projects.Include(p => p.PrincipalInvestigator).Where(p => p.IsActive).ToArrayAsync());
+            return Ok(await _dbContext.Projects
+                .Include(p => p.PrincipalInvestigator)
+                .Where(p => p.IsActive && (
+                    //PI's can only view their own projects
```
srkirkland, line 48: I think a better way to do this would be to have a separate method for Active() and Mine().  One shows all projects you can see, one just shows where you are PI.  I think that's how Cal wanted it originally but we should think about it.  Also, please do some parenthesis when doing multiple conditionals like this.

Harvest.Web/Startup.cs
```diff
@@ -86,17 +86,33 @@ public void ConfigureServices(IServiceCollection services)
...
+                    // Supervisor can access anything restricted to Supervisor, Worker or PI roles
+                    AccessCodes.SupervisorAccess => new[] { Role.Codes.Supervisor, Role.Codes.FieldManager },
+                    // Worker can access anything restricted to Worker role
+                    AccessCodes.WorkerAccess => new[] { Role.Codes.Worker, Role.Codes.Supervisor, Role.Codes.FieldManager },
+                    // PI can access anything restricted to PI role
+                    AccessCodes.PrincipalInvestigator => new[] { Role.Codes.PI, Role.Codes.Supervisor, Role.Codes.FieldManager },
+                    _ => throw new ArgumentException($"{nameof(accessCode)} is not a valid {nameof(AccessCodes)} constant")
+                };
+            }
+
```
srkirkland, line 107: This is pretty nice.  throw it in a helper file somewhere though so we can keep startup slimmer.

Harvest.Core/Services/UserService.cs
sprucely: I prefer keeping interfaces simple. It seems redundant to declare something in an interface when it can be implemented as a static method against the interface.

Harvest.Web/ClientApp/src/App.tsx
jSylvestre: Request or quote? FM should approve requests, PI the quote.

Harvest.Web/Controllers/ProjectController.cs
sprucely: Splitting into Active() and Mine() means moving the check on PI to client-side. The fetch effects for both ProjectListContainer and ProjectSelection components would need to be updated.

jSylvestre: @srkirkland Do you mean have the ProjectListContainer.tsx call a different endpoint based on roles?

srkirkland: Yes I agree

Harvest.Web/ClientApp/src/App.tsx
srkirkland: @jSylvestre I don't understand, what is the FM approving?  The only approval button is to approve the quote, which is /request/approve/ url

Harvest.Web/Controllers/ProjectController.cs
srkirkland: @sprucely @jSylvestre let me clarify better -- no, the container wouldn't call different endpoints based on role, and no checks would be moved client side.  We would just have two different routes.  You would need to be supervisor or FM to visit the regular see all projects list.  Anyone else could see /projects/me and it would show projects where they are a PI.

Harvest.Web/ClientApp/src/App.tsx
jSylvestre: Ah. Just me being confused. When you said approving requests, I was thinking project not quote

sprucely: Fixed

Harvest.Web/ClientApp/src/App.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/App.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/App.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/App.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
sprucely: Fixed

Harvest.Web/ClientApp/src/Shared/ShowFor.tsx
sprucely: Fixed

Harvest.Web/Controllers/ProjectController.cs
sprucely: done

Harvest.Web/Startup.cs
sprucely: done

sprucely:
The biggest change in my latest commit is adding PI to user roles if they are PI for the given projectId, or if there is no projectId in the request and the user is PI for at least one project.

Harvest.Web/ClientApp/src/Projects/ProjectListContainer.tsx
```diff
@@ -7,13 +7,17 @@ import { ProjectTable } from "./ProjectTable";
...
+}
+
+export const ProjectListContainer = (props: Props) => {
   const [projects, setProjects] = useState<Project[]>([]);
 
   useEffect(() => {
     // get rates so we can load up all expense types and info
     const cb = async () => {
-      const response = await fetch("/Project/Active");
+      const response = await fetch(props.projectSource);
```
jSylvestre, line 20: I was wondering how you were going to implement that. Cool.

srkirkland (approved):

PR #308, opened 2021-07-15
Package webpack outputs fancier

srkirkland:
Trying to mimic the optimized prod output from CRA.  Scripts now inlines the runtime-main JS bootstrapper, and then adds other JS chunks as needed, main last.  

CSS is pretty simple, we just sort so that higher numbers are first, then main is last.


jSylvestre (approved):
Seems ok to me.

PR #310, opened 2021-07-15
cache static spa assets

srkirkland:
Only happens in production -- locally we serve from webpack and so the static file handler doesn't apply.

Anything from the /static/ path (specifically anything from the SPA build folder served under the /static/ path) gets cached for a long time.  Since these are hashed it should be fine to do.


jSylvestre (approved):
When we deploy, these get new names so they will get updated, right?

srkirkland:
> When we deploy, these get new names so they will get updated, right?

That's the beauty -- they only get new names when the content changes (hashed).  So for 95% of changes if we don't add any NPM modules, the whole vendor file doesn't change and stays in cache.  If we just make a server-side change then none of the files will change.  But yes, if we modify some JS, the file name will change.

PR #311, opened 2021-07-15
adds a profile which allows building/running in prod mode locally

srkirkland:
useful for debugging.  doesn't affect any default behaviors.  run with `npm run production`

Useful to note that when running in prod your user-secrets don't load, since that's a dev only thing.

### DIFF
--- a/Harvest.Web/Properties/launchSettings.json
+++ b/Harvest.Web/Properties/launchSettings.json
@@ -22,6 +22,14 @@
         "ASPNETCORE_ENVIRONMENT": "Development"
       },
       "applicationUrl": "https://localhost:5001;http://localhost:5000"
+    },
+    "Harvest.Web.Production": {
+      "commandName": "Project",
+      "launchBrowser": true,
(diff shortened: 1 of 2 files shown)


Harvest.Web/Properties/launchSettings.json
```diff
@@ -22,6 +22,14 @@
...
       },
       "applicationUrl": "https://localhost:5001;http://localhost:5000"
+    },
+    "Harvest.Web.Production": {
+      "commandName": "Project",
+      "launchBrowser": true,
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Production"
+      },
+      "applicationUrl": "https://localhost:5001;http://localhost:5000"
```
jSylvestre, line 32: Prod is localhost?

srkirkland: on my machine it is ;0.  yea this is for testing prod builds locally.

jSylvestre (approved):

PR #312, opened 2021-07-15
Submit button disabled until text fields are inputted

### LINKED ISSUE #280: disabled button unless required is filled on creating new project

wongband:
resolve #280

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect, useMemo, useState } from "react";
 import { useHistory, useParams } from "react-router";
 import { Button, FormGroup, Input, Label } from "reactstrap";
 import { ValidationError } from "yup";
@@ -26,7 +26,6 @@ export const RequestContainer = () => {
     principalInvestigator: userDetail,
   } as Project);
   const [inputErrors, setInputErrors] = useState<string[]>([]);
-  const [buttonDisabled, setButtonDisabled] = useState<boolean>(true);
 
   const checkRequestValidity = async (inputs: any) => {
     try {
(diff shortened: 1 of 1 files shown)


Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
```diff
@@ -58,6 +59,18 @@ export const RequestContainer = () => {
...
+  useEffect(() => {
+    if (
+      project.start !== undefined &&
+      project.end !== undefined &&
+      project.crop !== undefined
+    ) {
+      setButtonDisabled(false);
+    }
+  }, [project.crop, project.start, project.end]);
+
```
srkirkland, line 73: Try this instead -- get rid of the state/boolean.  Change this to `useMemo`.  So 
```
var isFilledIn =  useMemo(() => {
    return
      project.start !== undefined &&
      project.end !== undefined &&
      project.crop !== undefined;
  }, [project.crop, project.start, project.end]);
```

then you just need to set button `disabled={!isFilledIn}`.  Save yourself some state and use less code for it.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
wongband: Okay I can update that. Also, when is it appropriate to use `useMemo` instead of `useEffect` in React?

srkirkland: UseMemo returns a value so it's useful when you want the result of some calculation locally. UseEffect is just something that runs based on changed params, so good for making network calls or updating totals or variables.

srkirkland (approved):

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
wongband: Ahh I see. So useMemo is mainly used when we don't want to cause side effects like making API calls while useEffect does with side effects?

srkirkland: Yea that's how I tend to use them anyway. I suppose you could have a side effect inside memo but that's probably bad form.

PR #313, opened 2021-07-16
Jcs/ticket notifications

jSylvestre:
image

### DIFF
--- a/Harvest.Web/Services/EmailService.cs
+++ b/Harvest.Web/Services/EmailService.cs
@@ -10,6 +10,7 @@ using Harvest.Email.Models;
 using Harvest.Email.Models.Ticket;
 using Harvest.Email.Services;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.IdentityModel.Tokens;
 using Serilog;
 
 namespace Harvest.Web.Services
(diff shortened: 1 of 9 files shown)


Harvest.Web/Services/EmailService.cs
```diff
@@ -25,13 +27,18 @@ public interface IEmailService
...
+        Task<bool> TicketAttachmentAdded(Project project, Ticket ticket, TicketAttachment[] ticketAttachments);
+        Task<bool> TicketClosed(Project project, Ticket ticket);
     }
 
     public class EmailService : IEmailService
     {
         private readonly AppDbContext _dbContext;
         private readonly IEmailBodyService _emailBodyService;
         private readonly INotificationService _notificationService;
+        public const string BaseUrl = "https://harvest.caes.ucdavis.edu";
```
sprucely, line 41: Maybe have this as an option, so that we can configure it in secrets to point to localhost?

sprucely (approved):

Harvest.Web/Services/EmailService.cs
jSylvestre: Yeah. Just being lazy for now.

srkirkland (commented):
Got a lot of commented code to delete but otherwise looks good.

PR #317, opened 2021-07-16
Don't fall through to SPA, fallback to ErrorController

sprucely:
Falling through to spa results in returning CRA's default index.html. This adds a fallback mapping to Error/Index, adding a statusCode value of 404. I wasn't able to test in a locally published folder in release mode because of some weird problem with React.createContext not being defined.


srkirkland (commented):
I don't understand why not just add a * route to react router which shows a 404 component?

sprucely:
It ends up loading CRA's default index.html file which has none of the content from _Layout.html.

srkirkland (approved):
Ok that's makes sense.

Issue #267, opened 2021-06-25
field map overlays


srkirkland:
layers for fields map.

1. Show fields for approved projects which will be in-use at the same time as project.
2. same but include proposed/non-approved projects as well.

clicking on a field from the layer should say which project, timeline and crop.

srkirkland:
Fields for approved projects has been added, we'll stop there for now.  Adding more will be pretty simple if other scenarios appear.

PR #323, opened 2021-07-19
show overlays of existing fields

srkirkland:
Any field that would be part of a project which overlaps temporally the project you are quoting for should show up so you know what is available.  Can be toggled on/off via leaflet layers control.

Does the bulk of #267 but we might want to make it fancier depending on what would be useful to the field managers.

### DIFF
--- a/Harvest.Web/ClientApp/src/Fields/FieldLayers.tsx
+++ b/Harvest.Web/ClientApp/src/Fields/FieldLayers.tsx
@@ -1,0 +1,52 @@
+import React, { useEffect, useState } from "react";
+import { LayersControl, GeoJSON, Popup, LayerGroup } from "react-leaflet";
+import { Field, Project } from "../types";
+
+interface Props {
+  project: Project;
+}
+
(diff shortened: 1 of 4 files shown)


Harvest.Web/ClientApp/src/Fields/FieldLayers.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const FieldLayers = (props: Props) => {
+  const [fields, setFields] = useState<Field[]>([]);
+
+  // load fields for active projects
+  useEffect(() => {
+    const cb = async () => {
+      const startDate = new Date(props.project.start);
+      const endDate = new Date(props.project.end);
```
jSylvestre, line 16: Just start and end dates? Don't want to include active projects?

srkirkland: @jSylvestre if you look at the C# controller it does filter by just approved projects.  They don't have to be currently active.

jSylvestre (approved):

PR #325, opened 2021-07-19
Showed accounts if they exist on the ProjectHeader

### LINKED ISSUE #319: Show account(s) that are used on the project
If there are any accounts, show them somewhere on the account details page and their percentage.
Perhaps in the Projectheader above the requirements.

wongband:
resolve #319

### DIFF
--- a/Harvest.Core/Domain/Project.cs
+++ b/Harvest.Core/Domain/Project.cs
@@ -85,7 +85,6 @@ namespace Harvest.Core.Domain
         [Display(Name = "Created By")]
         public User CreatedBy { get; set; }
 
-        [JsonIgnore]
         public List<Account> Accounts { get; set; }
 
         [JsonIgnore]
--- a/Harvest.Core/Domain/Account.cs
+++ b/Harvest.Core/Domain/Account.cs
@@ -1,8 +1,7 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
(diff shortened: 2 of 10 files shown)


jSylvestre:
@wongband FYI, if you use a div instead of a p tag, you can reduce the line spacing:

image
vs.
image

srkirkland (approved):
agree with jason it would be nicer w/ less spacing.  otherwise looks good to me.

Harvest.Core/Domain/Project.cs
```diff
@@ -85,7 +85,6 @@ public Project()
         [Display(Name = "Created By")]
         public User CreatedBy { get; set; }
 
-        [JsonIgnore]
```
jSylvestre, line 88: Don't think you want to remove this.

jSylvestre (changes requested):

Harvest.Core/Domain/Project.cs
jSylvestre: nm I'm probably wrong.

jSylvestre (approved):

Harvest.Core/Domain/Project.cs
wongband: Is there another way to get the data if we JsonIgnore it? I JsonIgnored the Project field in Account.cs to prevent circular referencing. I think I mentioned this in #sow-what

jSylvestre: I think you did it correctly.

PR #327, opened 2021-07-19
default project approval to false

### LINKED ISSUE #321: Projects should be default non-approved, then get approved only when approved by PI

srkirkland:
set approved to true only when the PI has approved it.

fixes #321


jSylvestre (approved):
Didn't test, but looks good.

PR #329, opened 2021-07-20
Jcs/fix create request

### LINKED ISSUE #324: Re-check errors on create project
Create a project, don't select crop.
Hit create
Add crop
Can't submit now:

image

jSylvestre:
resolves #324

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
@@ -109,7 +109,8 @@ export const RequestContainer = () => {
     return (
       project.start !== undefined &&
       project.end !== undefined &&
-      project.crop !== undefined
+      project.crop !== undefined &&
+      project.crop !== ""
     );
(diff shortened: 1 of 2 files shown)


Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
```diff
@@ -108,8 +108,11 @@ export const RequestContainer = () => {
   var isFilledIn = useMemo(() => {
     return (
       project.start !== undefined &&
+      project.start !== null &&
       project.end !== undefined &&
-      project.crop !== undefined
+      project.end !== null &&
+      project.crop !== undefined &&
+      project.crop !== ""
```
srkirkland, line 115: Probably better to just use falsey values here.  Usually strict equals is nice but if there are this many options i'd be easier to just say `project.start && project.end && project.crop` and be done with it.

srkirkland (approved):

wongband (approved):

PR #334, opened 2021-07-20
Jcs/ticket fixes20210720

jSylvestre:


srkirkland (approved):
This is probably fine -- it's probably more common to have a different endpoint to load messages, but we can live with this since it's simpler and we probably won't have tickets with tons of messages

PR #338, opened 2021-07-21
Try to get projectId from both Query and RouteValues

### LINKED ISSUE #333: PI access for posts
This method isn't getting the project id to auth the PI against:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Web/Controllers/Api/TicketController.cs#L143

sprucely:
Closes #333


jSylvestre (approved):
Works for me.

srkirkland (approved):

PR #339, opened 2021-07-21
Attachments can now be downloaded

### LINKED ISSUE #331: Ability to download files
- [x] Download from project page
- [x] Download ticket attachments

Download code should already be mostly written in the fileController, but you'll need to augment it to take in the file name.  It'll return a SAS URL which the client should use to download.  to save on network requests, you might want to add another method as well that gets the read SAS URL for multiple files at once.

wongband:
resolve #331

### DIFF
--- a/Harvest.Web/Controllers/Api/TicketController.cs
+++ b/Harvest.Web/Controllers/Api/TicketController.cs
@@ -137,14 +137,10 @@ namespace Harvest.Web.Controllers.Api
                         Name = a.Name, CreatedBy = a.CreatedBy, CreatedOn = a.CreatedOn, UpdatedBy = a.UpdatedBy,
                         Requirements = a.Requirements, WorkNotes = a.WorkNotes,
                         UpdatedOn = a.UpdatedOn, DueDate = a.DueDate, Status = a.Status, Messages =  a.Messages.Select(b => new TicketMessage{Id = b.Id, CreatedBy = b.CreatedBy, CreatedOn = b.CreatedOn, Message = b.Message}).ToList(),
-                        Attachments = a.Attachments.Select(b => new TicketAttachment{Id = b.Id, CreatedBy = b.CreatedBy, CreatedOn = b.CreatedOn, FileName = b.FileName, Identifier = b.Identifier}).ToList()
+                        Attachments = a.Attachments.Select(b => new TicketAttachment{Id = b.Id, CreatedBy = b.CreatedBy, CreatedOn = b.CreatedOn, FileName = b.FileName, Identifier = b.Identifier, SasLink = fileService.GetDownloadUrl(storageSettings.ContainerName, b.Identifier).AbsoluteUri}).ToList(),
                     })
                 .SingleAsync();
             
-            foreach(var file in ticket.Attachments){
-                file.SasLink = fileService.GetDownloadUrl(storageSettings.ContainerName, file.Identifier).AbsoluteUri;
-            }
-
             return Ok(ticket);
         }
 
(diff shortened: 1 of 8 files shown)


srkirkland (approved):

sprucely (approved):
Works for me.

It might be worthwhile to add `attachment; ` to the beginning of the content-disposition string so that clicking the link will result in a save-as dialog:
https://github.com/ucdavis/Harvest/blob/4ca315b98ab82f025ad7e641791149901a0b0176/Harvest.Web/ClientApp/src/Shared/FileUpload.tsx#L30

jSylvestre:
@wongband Yes, please make it so it saves the attachment instead of opening it on the page.

wongband:
I will change that right away

srkirkland:
I believe you can just add `download` to the `<a>` tag and that'll do it

srkirkland:
@wongband while @sprucely would work, I specifically didn't want to make the blob storage hardcoded to an attachment -- we don't know if that's how it should be used in all cases.  Try the download attribute first, and if that works it'll be the most flexible option.  If not then we can go with the other one.

wongband:
@srkirkland download attribute does not work. Do i just changed `blobContentDisposition: `filename=${file.fileName}`` to  blobContentDisposition: `attachment; filename=${file.fileName}`,

srkirkland:
Thanks for checking, not sure why it doesn't work but yea just add that attachment; like you put there

jSylvestre:
If we don't go as a download, then at least open it up in a new tab in the browser.

wongband:
When I change it to `attachment; filename=${file.fileName}` and I click on the file it does not download it still opens the attachment

sprucely:
Did you try uploading a new file? It's metadata that gets saved with the blob, so previous uploads will not have it.

jSylvestre:
Or just add this to the a tag target="_blank" rel="noreferrer"

wongband:
@sprucely I think we want all the attachments to be downloadable and not just the the uploaded ones

sprucely:
@wongband It's not in production yet, so with that `attachment;` disposition being applied from the beginning, they would all be downloadable.

jSylvestre:
Yeah, adding that to the fileupload.tsx worked for me.

wongband:
@sprucely ohh okay I see. Should I keep the attachment there or do what @jSylvestre with the blank? That method works as well

jSylvestre:
@wongband I'd say just go with the blank for now to it opens in a new tab. If we decide we want to force a download for some of these in the future we can modify the upload for those cases.

wongband:
Adding the blank to the `a` tag allows the document to be downloaded. Is this fine as well?

Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -135,6 +140,11 @@ public async Task<ActionResult> Get(int projectId, int ticketId)
                         Attachments = a.Attachments.Select(b => new TicketAttachment{Id = b.Id, CreatedBy = b.CreatedBy, CreatedOn = b.CreatedOn, FileName = b.FileName, Identifier = b.Identifier}).ToList()
```
jSylvestre, line 140: Instead of doing the foreach below, I think you can do it in here:
....... , SasLink = fileService.GetDownloadUrl(storageSettings.ContainerName, b.Identifier).AbsoluteUri }).ToList()

jSylvestre (approved):
FYI, for the project attachments, if you upload more than 1, it isn't downloadable unless you refresh. In this example, there was 1 attachment, then I uploaded 2 more:

image

wongband:
@jSylvestre check it again. It should be fixed with that last push

PR #342, opened 2021-07-22
Add PI to Harvest.Roles when user is PI for any projects

sprucely:


srkirkland (approved):

srkirkland (commented):
Maybe consider toList for easier appending

Issue #201, opened 2021-06-02
History: Add history for all important actions in the app


srkirkland:
Some that I can think of that need to be implemented.  Add more and check off as needed:

- [x] Request created
- [x] Quote submitted
- [x] Quote Approval
- [x] Account(s) changed
- [x] Project completed
- [x] Ticket created
- [ ] Ticket completed - currently no business logic for marking ticket as completed
- [x] Expense created
- [x] Expense deleted

Minor stuff which might need history

- [x] Quote save/edit
- [x] File(s) ~~uploaded?~~ attached (actual upload is performed client-side)
- [x] Ticket messages?
- [x] Ticket work notes updated

srkirkland:
ok let's call this done for now then

PR #343, opened 2021-07-22
Swe/add history

sprucely:
Addresses most of #201

It was hard to decide what to include in each history model. They are kind of a mishmash of near-complete mirrors of domain entities, small models with just enough data to be relevant to a given action, and already-existing models. 

A lot of the serialized history has pk Id values of 0, since the dbContext has not been saved at the time of serialization. Should I place the history generation after dbContext saves and then perform a second save just for the history?

### DIFF
--- a/Harvest.Core/Models/History/RateHistoryModel.cs
+++ b/Harvest.Core/Models/History/RateHistoryModel.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Harvest.Core.Domain;
+
+namespace Harvest.Core.Models.History
+{
+    public class RateHistoryModel
+    {
+        public static RateHistoryModel CreateFrom(Rate rate)
+        {
+            if (rate == null)
+            {
+                return null;
+            }
+
+            return new RateHistoryModel
+            {
+                Id = rate.Id,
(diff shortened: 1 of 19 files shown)


Harvest.Core/Models/History/RateHistoryModel.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Harvest.Core.Domain;
+
+namespace Harvest.Core.Models.History
+{
+    public class RateHistoryModel
+    {
+        public static RateHistoryModel CreateFrom(Rate rate)
```
srkirkland, line 12: why sometimes a static and sometimes you use the ctor?  Like `RateHistoryModel.CreateFrom(xyz)` vs sometimes you do `new RateHistoryModel(xyz)`?

srkirkland (commented):
Cool I like the direction this is going

Harvest.Core/Models/History/RateHistoryModel.cs
sprucely: For some, like User, they will only ever be instantiated from a property of a parent domain entity. Since those could be null based on whether they were included, I thought it would look cleaner than a bunch of ternary operators.

srkirkland: makes sense

PR #344, opened 2021-07-23
close ticket (Just updates status for now)

jSylvestre:


Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
```diff
@@ -62,25 +66,30 @@ export const TicketAttachments = (props: Props) => {
...
-              ] = { ...f };
+            }}
+            updateFile={(f) =>
+              setTicketLoc((ticket) => {
+                // update just one specific file from ticket p
+                ticket.newAttachments[
+                  ticket.newAttachments.findIndex(
+                    (file) => file.identifier === f.identifier
+                  )
+                ] = { ...f };
```
srkirkland, line 83: this doesn't quite seem right to me... usually in a set hook you are supposed to return the new value, not just modify the object.

jSylvestre: Not sure how to change that. But it is working.

PR #376, opened 2021-07-27
Logout and endemulation

jSylvestre:


Harvest.Web/Startup.cs
```diff
@@ -144,6 +145,7 @@ public void ConfigureServices(IServiceCollection services)
             services.Configure<SparkpostSettings>(Configuration.GetSection("SparkPost"));
             services.Configure<StorageSettings>(Configuration.GetSection("Storage"));
             services.Configure<EmailSettings>(Configuration.GetSection("Email"));
+            services.Configure<AuthenticationSettings>(Configuration.GetSection("Authentication"));
```
srkirkland, line 148: We should use the already existing auth settings file from a few lines up, the one that has cas login info.

srkirkland (changes requested):

jSylvestre:
Odd though... If I logout, then paste the local harvest URL in again, I get a CAS error.

srkirkland (approved):

PR #378, opened 2021-07-28
Toast library and example toast

srkirkland:
Thought i would get this out early, and if we like it I can start adding it into all components.  This library is pretty cool because we can pass a promise and it'll handle the spinner and updating, in addition to normal notifications.

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
@@ -15,7 +15,7 @@ import { ProjectHeader } from "../Shared/ProjectHeader";
 import { ActivitiesContainer } from "./ActivitiesContainer";
 import { QuoteTotals } from "./QuoteTotals";
 
-import { toast } from "../Util/Notifications";
+import { fetchWithFailOnNotOk, genericError, toast } from "../Util/Notifications";
 
 interface RouteParams {
   projectId?: string;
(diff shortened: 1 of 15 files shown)


sprucely (approved):
I like it. Do you think it would be helpful to wrap in a hook that returns something like `{ isLoading: bool, isSuccess: bool, error: any }`?

srkirkland:
> I like it. Do you think it would be helpful to wrap in a hook that returns something like `{ isLoading: bool, isSuccess: bool, error: any }`?

Interesting idea, not sure how to actually build it.  I'll keep it in mind to come back to after getting this lib integrated.

jSylvestre (approved):
I like the look more than what we have for Peaks

Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
```diff
@@ -150,10 +151,17 @@ export const QuoteContainer = () => {
       body: JSON.stringify(quote),
     });
 
+    toast.promise(request, {
+      loading: "Saving Quote",
+      success: "Quote saved.",
+      error: "Something went wrong, please try again",
```
jSylvestre, line 157: @srkirkland I tried to throw an exception and return a bad request. Neither displayed the error.

srkirkland:
closes #383

sprucely (approved):
Looks good, and BadRequests are giving toast errors.

jSylvestre (approved):
If you want to test the error, go to a ticket details page an reply with no text.

PR #381, opened 2021-07-29
Disable buttons while submitting

### LINKED ISSUE #371: disable submit button on approve page while approval is happening

jSylvestre:
Also disable reject which doesn't do anything yet.
resolves #371


srkirkland (commented):
Maybe hold off on this for now until we think of a holistic solution for all the save buttons.  Right now disabled isn't even ever being set so no reason to have multiple variables either.

jSylvestre:
> Maybe hold off on this for now until we think of a holistic solution for all the save buttons. Right now disabled isn't even ever being set so no reason to have multiple variables either.

@srkirkland FYI, disabled is set in the accounts container (need account and must add up to 100%)
But I'm fine waiting.

jSylvestre:
Re-visit later

PR #387, opened 2021-07-29
Fix closed ticket logic

jSylvestre:


Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
```diff
@@ -65,8 +65,8 @@ export const TicketAttachments = (props: Props) => {
         ))}
       </ul>
       <FormGroup>
-        <Label>Attach files?</Label>
-        {ticket.completed !== false ? (
+        {!ticket.completed && <Label>Attach files?</Label>}
+        {!ticket.completed && (
           <FileUpload
```
srkirkland, line 70: There is no reason to put these in two different lines -- they are the same conditional.  Just stick the attach files label inside and it should be fine.  might have to add an empty <> or a div but otherwise it should be fine.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Tickets/TicketAttachments.tsx
jSylvestre: I keep forgetting about the empty <>

srkirkland (approved):

PR #388, opened 2021-07-29
Call RequestCreated after request is created

### LINKED ISSUE #379: History service when creating a new project
Save fails:

The INSERT statement conflicted with the FOREIGN KEY constraint "FK_ProjectHistory_Projects_ProjectId". The conflict occurred in database "Harvest", table "dbo.Projects", column 'Id'.
The statement has been terminated.

sprucely:
closes #379


jSylvestre:
@sprucely I thought @srkirkland didn't like this solution....

srkirkland:
Can't we associate newProject with the history entry in the db and let EF handle the save. Saving twice like this feels like it isn't using EF correctly.

jSylvestre:
Some example code incoming

sprucely:
@srkirkland Now I get it. I'm so used to doing things with PK's that I forget about using navigation properties...

jSylvestre:
Done in other pr...

PR #389, opened 2021-07-29
Pass project not Id.

jSylvestre:
I'm sure this could be cleaned up


jSylvestre:
Possible solution to #379

sprucely (approved):
Haha. This is what I was going to do. But why name it MakeHistory2 instead of just overloading?

jSylvestre:
> Haha. This is what I was going to do. But why name it MakeHistory2 instead of just overloading?

I didn't see where it was failing originally and was just testing. I'll fix it up.

jSylvestre:
resolves #379

sprucely (approved):

PR #390, opened 2021-07-29
custom react hook for promise notification with status state

srkirkland:
Hook which can fire a toast message given a promise and which also returns state that can be inspected to determine promise status.

Might be part of solution for #371


jSylvestre (approved):
Didn't test, but looks good

PR #394, opened 2021-07-30
Swe/validation2.0

sprucely:
Created something much that much more closely suits our needs than formik and react-hook-form. It is strongly typed on a given data type. It uses yup's `validateAt()` against a temporary object with just the given property so that it doesn't have to concern itself with the entire object backing a controlled form. One shortcoming of this approach is that it will not work with yup validations that take into account more than one property, but I'm sure we could figure out a solution to that if the need ever arises.

[update]
Added ValidationProvider to enable returning some form-level stuff, (formIsTouched, formIsDirty, formErrorCount, resetContext).
Finished implementing on QuoteContainer.


srkirkland (approved):

srkirkland (approved):
nice i like it.

jSylvestre (changes requested):
Either disable the save when the form is validated as well as the submit, or validate the form when it is loaded if there is an existing quote...

PR #395, opened 2021-08-02
Create ticket validation

### LINKED ISSUE #360: Add validation to ticket create

jSylvestre:
resolves #360

### DIFF
--- a/Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
+++ b/Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
@@ -6,6 +6,8 @@ import DatePicker from "react-date-picker";
 import { Button, FormGroup, Input, Label } from "reactstrap";
 import { FileUpload } from "../Shared/FileUpload";
 import { ShowFor } from "../Shared/ShowFor";
+import { ticketSchema } from "../schemas";
+import { ValidationError } from "yup";
 import {
   fetchWithFailOnNotOk,
   genericErrorMessage,
@@ -18,6 +20,7 @@ interface RouteParams {
 
 export const TicketCreate = () => {
   const { projectId } = useParams<RouteParams>();
+  const [inputErrors, setInputErrors] = useState<string[]>([]);
   const [project, setProject] = useState<Project>();
   const [ticket, setTicket] = useState<Ticket>({
     requirements: "",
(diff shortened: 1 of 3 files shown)


Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
```diff
@@ -42,8 +45,30 @@ export const TicketCreate = () => {
...
+    }
+  };
+
   const create = async () => {
-    // TODO: validation
+
+    const ticketErrors = await checkTicketValidity(ticket);
+
+    if (ticketErrors) {
+      if (ticketErrors.length > 0) {
```
sprucely, line 65: I believe you can do null conditionals just like C#: `if (ticketErrors?.length > 0)`

Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
```diff
@@ -146,6 +171,15 @@ export const TicketCreate = () => {
                   }
                 ></FileUpload>
               </FormGroup>
+              <ul>
+            {inputErrors.map((error, i) => {
+              return (
+                <li style={{ color: "red" }} key={`error-${i}`}>
```
sprucely, line 177: There is a css class `text-danger` that can be used for this.

sprucely (approved):
Looks okay. Left a couple comments, but didn't test.

Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
```diff
@@ -42,8 +45,30 @@ export const TicketCreate = () => {
...
+    const ticketErrors = await checkTicketValidity(ticket);
+
+    if (ticketErrors) {
+      if (ticketErrors?.length > 0) {
+        setInputErrors(ticketErrors);
+        return;
+      } else {
+        setInputErrors([]);
+      }
+    }
```
srkirkland, line 71: Just combine this into one, you don't need to nest here.  `if (ticketErrors?.length > 0)` is sufficient, or `if (ticketErrors && ticketErrors.length > 0)` if you like that better.

srkirkland (approved):

Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
jSylvestre: Argument of type 'string[] | undefined' is not assignable to parameter of type 'SetStateAction<string[]>'.
  Type 'undefined' is not assignable to type 'SetStateAction<string[]>'.

srkirkland: I'm confused here, how are you getting that error if you do `if (ticketErrors && ticketErrors.length > 0)`, that should guarentee it can't be undefined.  what line are you getting it on.

Harvest.Web/ClientApp/src/Tickets/TicketCreate.tsx
```diff
@@ -40,8 +43,30 @@ export const TicketCreate = () => {
...
+
   const create = async () => {
-    // TODO: validation
+
+    const ticketErrors = await checkTicketValidity(ticket);
+
+    if (ticketErrors) {
+      if (ticketErrors.length > 0) {
+        setInputErrors(ticketErrors);
+        return;
```
jSylvestre, line 65: If there were any errors, it returns early

PR #398, opened 2021-08-02
Ability to reject quote

### LINKED ISSUE #370: Reject button on approve page
Should pop open modal, require reject reason, create ticket based on that.

Puts project into new status of "QuoteRejected".

srkirkland:
new status, reject creates a ticket with reject reason

closes #370

### DIFF
--- a/Harvest.Web/Controllers/Api/RequestController.cs
+++ b/Harvest.Web/Controllers/Api/RequestController.cs
@@ -129,6 +129,21 @@ namespace Harvest.Web.Controllers
 
         [HttpPost]
         [Authorize(Policy = AccessCodes.PrincipalInvestigator)]
+        public async Task<ActionResult> RejectQuote(int projectId, [FromBody] string reason) {
+            var project = await _dbContext.Projects.SingleAsync(p => p.Id == projectId);
+
+            project.Status = Project.Statuses.Requested;  // TODO: do we want a rejected status?
+
+            // TODO: history for rejected quote
+            // TODO: email for rejected quote including reason
+
+            await _dbContext.SaveChangesAsync();
+
+            return Ok(project);
+        }
+
+        [HttpPost]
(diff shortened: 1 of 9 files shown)


Harvest.Web/Controllers/Api/RequestController.cs
```diff
@@ -119,14 +119,44 @@ public async Task<ActionResult> Approve(int projectId, [FromBody] RequestApprova
...
+            await _dbContext.Tickets.AddAsync(ticketToCreate);
+            await _historyService.TicketCreated(project.Id, ticketToCreate);
+
+            project.Status = Project.Statuses.QuoteRejected;
+
+            await _historyService.QuoteRejected(project.Id, model.Reason);
+
+            await _dbContext.SaveChangesAsync();
+
+            await _emailService.NewTicketCreated(project, ticketToCreate);
```
jSylvestre, line 155: @srkirkland There should be a quote rejected endpoint.

srkirkland: @jSylvestre should i do that instead of doing a ticket even though a ticket's created?  It would probably make more sense than doing both i suppose.

srkirkland: @jSylvestre Let's do that in a future issue -- if we move over to that method we should definitely (a) show the actual reason it was rejected and (b) probably link to the project page or ticket for it.

jSylvestre: Up to you. Possibly I can change the quote rejected email to have a link to the ticket?
Or just don't use the quote rejected, but it does stand out as a significant email.

jSylvestre:
Change the "Approve Quote" button text on the project details to something like "Review Quote" ?

jSylvestre (approved):

PR #400, opened 2021-08-02
Added jest tests for ProjectDetailsContainer

### LINKED ISSUE #348: TESTING: ProjectDetailContainer
- [x] Check that basic header info (like notes or crop) are loaded correctly
- [x] Check recent invoices will show if they exist
- [x] Check recent tickets show if they exist
- [x] Project attachments show if they exist

wongband:
resolve #348

### DIFF
--- a/Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.test.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.test.tsx
@@ -4,9 +4,30 @@ import { MemoryRouter, Route } from "react-router-dom";
 import { act } from "react-dom/test-utils";
 
 import { ProjectDetailContainer } from "./ProjectDetailContainer";
-import { fakeAppContext, fakeProject } from "../Test/mockData";
+import {
+  fakeAppContext,
+  fakeInvoices,
+  fakeProject,
+  fakeTickets,
+} from "../Test/mockData";
 
 let container: Element;
+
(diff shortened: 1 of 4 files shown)


srkirkland (changes requested):
@wongband I think something isn't mocked correctly, I'm getting a fetch error when I run the tests `npm test`.  The build broke too for the same reason.  Seems to be in "Display correct number of recent tickets"

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+      global.fetch = jest
+        .fn()
+        .mockImplementationOnce(() => Promise.resolve(projectResponse))
+        .mockImplementationOnce(() => Promise.resolve(unbilledResponse))
+        .mockImplementationOnce(() => Promise.resolve(ticketResponses))
+        .mockImplementationOnce(() => Promise.resolve(invoiceResponse))
+        .mockImplementationOnce(() => Promise.resolve(fileResponse));
+
+      render(
+        <MemoryRouter initialEntries={["/project/details/3"]}>
```
jSylvestre, line 176: @wongband Does this need <AppContext.Provider value={(global as any).Harvest}>

wongband: I will add it in

PR #402, opened 2021-08-02
homepage is now a full react homepage

### LINKED ISSUE #399: Homepage in react
Replace the default generic server homepage with a homepage in react.  For starters we can just replicate but eventually this will help make it dynamic and allow our client routing to work properly.

srkirkland:
Now homepage can be routed via react router but also works when hit directly.  Can further use roles to customize quick links or whatever we want.

closes #399


jSylvestre (approved):
Just have to remember to update these when we have real links.

PR #407, opened 2021-08-03
Implementation of change requests

### LINKED ISSUE #374: Create quote for a change requested project should use a copy of the approved quote from original

srkirkland:
Creating a change request for an active project creates an entirely new project.  This PR copies over important project details to that newly created project, including the previously approved quote so there is a good starting point.

More importantly, this PR adds a lot of logic so when the change request is approved, change request disappears and the original project is updated with the new changed info.

closes #374


jSylvestre (approved):
Seems OK, but I thought call said the request changes page should have the project head of the original?

PR #411, opened 2021-08-04
add in project head just for change requests

### LINKED ISSUE #373: Add project header to create request page if it is a change request
So we can reference initial request

srkirkland:
closes #373


jSylvestre (approved):
I think that looks good:

image

PR #423, opened 2021-08-06
Expense validation and misc fixes

### LINKED ISSUE #96: Prevent Negative values
This is probably validation in general, but here is one example:

image

sprucely:
Closes #96


jSylvestre (approved):
It would be nice if we could have some kind of warning if the have partially did something.
But this is good enough for me:
image

PR #433, opened 2021-08-11
Added new page to display all field projects

### LINKED ISSUE #427: New page which shows all active project fields on a map
Very similar to the `<Location/>` component used in ProjectDetail.tsx, We want a page with a large map which shows all of the active fields.  If you click on a field it should show a popup with field detail, like the name and what crop is grown there and quick links to enter project expenses or view the project details.

Make up a url for it, say /project/map or something.

wongband:
resolve #427

### DIFF
--- a/Harvest.Web/ClientApp/src/App.tsx
+++ b/Harvest.Web/ClientApp/src/App.tsx
@@ -22,7 +22,6 @@ import { InvoiceListContainer } from "./Invoices/InvoiceListContainer";
 import { TicketCreate } from "./Tickets/TicketCreate";
 import { TicketsContainer } from "./Tickets/TicketsContainer";
 import { TicketDetailContainer } from "./Tickets/TicketDetailContainer";
-import { Map } from "./Maps/Map";
 
 // Global variable containing top-level app settings and info
 declare var Harvest: AppContextShape;
--- a/Harvest.Web/ClientApp/src/App.tsx
+++ b/Harvest.Web/ClientApp/src/App.tsx
@@ -100,7 +100,12 @@ function App() {
         path="/expense/unbilled/:projectId"
         component={UnbilledExpensesContainer}
       />
-      <Route exact path="/project/map" component={ProjectFields}/>
+      <ConditionalRoute
+        roles={["FieldManager", "Supervisor"]}
+        exact
+        path="/project/map"
+        component={ProjectFields}
+      />
     </AppContext.Provider>
   );
 }
--- a/Harvest.Web/ClientApp/src/App.tsx
+++ b/Harvest.Web/ClientApp/src/App.tsx
@@ -15,6 +15,7 @@ import { RequestContainer } from "./Requests/RequestContainer";
 import { AccountChangeContainer } from "./Requests/AccountChangeContainer";
 import { QuoteContainer } from "./Quotes/QuoteContainer";
 import { ProjectDetailContainer } from "./Projects/ProjectDetailContainer";
+import { ProjectFields } from "./Projects/ProjectFields";
 import { ProjectListContainer } from "./Projects/ProjectListContainer";
 import { InvoiceDetailContainer } from "./Invoices/InvoiceDetailContainer";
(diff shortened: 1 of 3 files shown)


Harvest.Web/Controllers/ProjectController.cs
```diff
@@ -147,6 +152,15 @@ public async Task<ActionResult> Get(int projectId)
             return Ok(project);
         }
 
+        [HttpGet]
+        [Authorize(Policy = AccessCodes.FieldManagerAccess)]
+        public async Task<ActionResult> GetFields()
+        {
+            var fields = await _dbContext.Fields.ToListAsync();
```
sprucely, line 159: Should this be limited to fields marked IsActive and/or whose projects are active?

jSylvestre:
@wongband This is no longer used: `import { Map } from "./Maps/Map";` in the app.tsx

Harvest.Web/Controllers/ProjectController.cs
wongband: That is a good point. I will change that

wongband:
@jSylvestre I will remove that.

Harvest.Web/Controllers/ProjectController.cs
sprucely: Oh, also I think it should be authorized with `AccessCodes.WorkerAccess`.

Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -100,7 +100,7 @@ function App() {
         path="/expense/unbilled/:projectId"
         component={UnbilledExpensesContainer}
       />
-      <Route path="/home/map" component={Map} />
+      <Route exact path="/project/map" component={ProjectFields}/>
```
jSylvestre, line 103: I think this should be using the conditional route. Something like this
`      <ConditionalRoute
        exact
        roles={["FieldManager"]}`

jSylvestre: Based on Spruce's other reply. Worker, FM, and supervisor... Not sure what that looks like.

wongband: It's possible to add multiple authorize policies to one route correct?

Harvest.Web/Controllers/ProjectController.cs
sprucely: @wongband It looks like new fields default to being marked IsActive, and there's no logic that changes it to false, so be sure to include `field.Project.IsActive && field.Project.Status == Project.Statuses.Active` in the filter.

Harvest.Web/ClientApp/src/App.tsx
sprucely: @jSylvestre Do you think it should be `"FieldManager"`? I was thinking `"Worker"`

Harvest.Web/Controllers/ProjectController.cs
wongband: @sprucely I noticed that all of the fields' Project field is null. Is it suppose to be like this?

Harvest.Web/ClientApp/src/App.tsx
srkirkland: I think just Supervisor/FieldManager

srkirkland: Later we might re-use the basics of this component for a worker facing page but doesn't need to be this one

Harvest.Web/Controllers/ProjectController.cs
srkirkland: @wongband It'll work in the query.  The field is null when you get the result from the database because it won't pull relations by default.  But it'll work from the query itself since it just gets transformed to SQL.  If you really do need the project full data you can do a `.Include()` but I don't think you need that for this case.

jSylvestre (approved):

jSylvestre:
Works for me. 🚢 

image

wongband:
Is it good to merge?

jSylvestre:
@wongband Yes.

PR #435, opened 2021-08-12
Add parts markup for "other" category

### LINKED ISSUE #160: Other category items can specify "parts markup"
Every item in the other category has an optional parts markup, defaulting to false.  If it is selected, it'll raise the total price of the item by 20%.  

Detail from Brian M:
>  For pass through costs (irrigation supplies, fertilizers, etc.) our policy/practice in Russell Ranch is to add a 20% mark up to the pass through of any supplies, parts, or materials (the markup covers our costs acquiring, handling, and accounting for materials).  On the template we program this with a Yes/No dropdown selector.  But in the app I was thinking that there could be a check box for “Parts Markup”.  If the person entering the expenses checks the box, the total for that row/item would update to the entered total plus 20%.

srkirkland:
DB Change -- new column for tracking if there was a markup changed for a given expense.  Quote also updated.  If this new markup checkbox is checked it'll add 20% to the cost of the item.

I'll have Cal update the styling a bit more before I merge too.

closes #160

### DIFF
--- a/Harvest.Core/Domain/Expense.cs
+++ b/Harvest.Core/Domain/Expense.cs
@@ -27,6 +27,8 @@ namespace Harvest.Core.Domain
         [Range(0.01, Double.MaxValue, ErrorMessage = "Quantity must be greater than 0.01")]
         public decimal Quantity { get; set; }
 
+        public bool Markup { get; set; }
+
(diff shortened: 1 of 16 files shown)


srkirkland:
Styles done, ready for review

jSylvestre (approved):
Just don't like that the off looks like it is disabled.
But works for me.

PR #440, opened 2021-08-12
Get invoice job working

sprucely:


Harvest.Core/Services/InvoiceService.cs
```diff
@@ -139,14 +141,14 @@ private async Task CreateMonthlyAcreageExpense(Project project)
 
         }
 
-        public async Task<int> CreateInvoices()
+        public async Task<int> CreateInvoices(bool manualOverride = false)
         {
-            var activeProjects = await _dbContext.Projects.Where(a => a.IsActive && a.Status == Project.Statuses.Active)
+            var activeProjects = await _dbContext.Projects.Where(a => a.IsActive && a.Status == Project.Statuses.Active && a.Accounts.Any())
```
jSylvestre, line 146: If the project is active, but there are no accounts I think we should throw an exception.

Harvest.Core/Services/SlothService.cs
```diff
@@ -59,8 +60,12 @@ public async Task<SlothResponseModel> MoveMoney(int invoiceId)
                 .Include(a => a.Project).ThenInclude(a => a.Accounts).SingleOrDefaultAsync();
             if (invoice == null)
             {
-                Log.Error("Invoice not found: {invoiceId}", invoiceId);
-                return null;
+                return Result.Error("Invoice not found: {invoiceId}", invoiceId);
+            }
+
+            if (invoice.Project.Accounts.Count == 0)
```
jSylvestre, line 66: Can't get to this currently?

Harvest.Core/Models/Result.cs
```diff
@@ -21,7 +27,50 @@ public class Result<T>
     // the generic argument. So adding a bit of type inference and implicit casting makes for a better experience without getting too fancy...
     public static class Result
     {
-        public static ResultError Error(string errorMessage) => new ResultError(errorMessage);
+        public static ResultError Error(string messageTemplate, 
+            LogEventLevel logLevel = LogEventLevel.Error,
+            [CallerFilePath] string callerFilePath = "", 
+            [CallerLineNumber] int callerLineNumber = 0)
+        {
+            using var _ = LogContext.PushProperty("FileName", Path.GetFileName(callerFilePath));
```
sprucely, line 35: Not pretty, but the _ variable is disposed after function exits, removing FileName and LineNumber properties from log context.

Harvest.Core/Services/SlothService.cs
```diff
@@ -127,8 +142,7 @@ public async Task<SlothResponseModel> MoveMoney(int invoiceId)
                 var credit = await _financialService.IsValid(expense.Key);
```
jSylvestre, line 142: If an account isn't valid and it is preventing processing, maybe we should throw off an email too.

jSylvestre (approved):
Didn't test, and just a few comments to think about.

Harvest.Core/Services/SlothService.cs
srkirkland: yea agree with this, otherwise there'll be no way to find it.  not sure the best way though because we don't want to send out emails every day

srkirkland (approved):

Harvest.Core/Services/SlothService.cs
jSylvestre: It is possible that the validation will fail because of KFS issues.
Maybe trust them here, but have a job that validates accounts once a week or so? @srkirkland

srkirkland: don't really love an update job idea but maybe instead we just trust them always since Brian's team is doing billing review anyway?  And he can note when accounts need to be changed?  Or alternately we can do an invoice status of accountInvalid and then put up some big warning if there are any in that state. Maybe create a ticket automatically if any accounts are invalid

jSylvestre: Yeah, probably just let sloth/Brian deal with it.

Harvest.Core/Services/SlothService.cs
sprucely: It did occur, though it was probably old data that won't happen again with current checks in place.

PR #441, opened 2021-08-12
Fixed file upload error

### LINKED ISSUE #386: Files can't be downloaded after being uploaded
low priority but when i'm on the project details page and I upload a file, it shows up in the list but you can't download it.  If there was an existing file you can download that one and you are left with one downloadable and one not.  A page refresh makes the file downloadable again.

wongband:
resolve #386


srkirkland (approved):
seems to work for me

jSylvestre (approved):
Looks like it works to me, but maybe wait for Scott to take a look.

PR #446, opened 2021-08-17
Added cleanup function for useEffect to prevent error

### LINKED ISSUE #420: Console error when creating expenses
It doesn't cause any problems I can see, but when you create an expense and submit it, I get this in the console:

```
index.js:1 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at ExpenseEntryContainer (https://localhost:5001/static/js/main.chunk.js:737:86)
```

I'm thinking it's related to the way we do history.push instead of using the ReactRouter Redirect component but that's just a wild guess.

wongband:
resolve #420

### DIFF
--- a/Harvest.Web/ClientApp/src/Expenses/ExpenseEntryContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Expenses/ExpenseEntryContainer.tsx
@@ -50,6 +50,8 @@ export const ExpenseEntryContainer = () => {
   const [notification, setNotification] = usePromiseNotification();
 
   useEffect(() => {
+    let isMounted = true;
+
     // get rates so we can load up all expense types and info
     const cb = async () => {
       const response = await fetch("/Rate/Active");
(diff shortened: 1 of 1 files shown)


jSylvestre (approved):

srkirkland (approved):
I like it, but also interesting that I've only ever seen the issue here.  I would think since dependencies are `[]` it should only ever run once anyway...

wongband:
@srkirkland https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup. The documentation says that:
```
React performs the cleanup when the component unmounts. However, as we learned earlier, effects run for every render and 

not just once. This is why React also cleans up effects from the previous render before running the effects next time. We’ll 

discuss why this helps avoid bugs and how to opt out of this behavior in case it creates performance issues later below.
```
Strange though since I would expect other components to have something similar to this...

srkirkland:
@wongband except we pass an empty array, so it shouldn't run on every render:
> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

from https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects.  That's why I wouldn't think it should try to run again.

wongband:
@srkirkland ohh that's true. Maybe something else was leaking from the useEffect after it initially loads?

PR #449, opened 2021-08-18
Don't let invoice exceed quoted amount

### LINKED ISSUE #159: Don't create/process invoice if total project paid $$ is over quoted amount

sprucely:
resolves #159

There was too much aspnetcore infrastructure that would have needed to be manually configured to get `IRazorViewEngine` working in a console app, so I replaced EmailBodyService with Razor.Templating.Core.


Harvest.Core/Extensions/DateTimeExtensions.cs
```diff
@@ -18,6 +19,29 @@ private static TimeZoneInfo GetPacificTimeZone()
...
+
+        public static bool IsFirstBusinessDayOfMonth(this DateTime dateTime)
+        {
+            var pacificTime = dateTime.ToPacificTime();
+            var day = pacificTime.DayOfWeek;
+            return day switch
+            {
+                DayOfWeek.Saturday => false,
+                DayOfWeek.Sunday => false,
+                DayOfWeek.Monday => pacificTime.Day <= 3,
```
jSylvestre, line 40: I don't like this because it isn't clear what it is doing. But maybe I'm just too old.

Harvest.Core/Services/InvoiceService.cs
```diff
@@ -25,57 +25,74 @@ public class InvoiceService : IInvoiceService
...
-            if (project.End < now)
-            {
-                //Ok, so the project has ended, we should probably create the last invoice, and close it out. Setting the project to completed.
-                project.Status = Project.Statuses.Completed;
-            }
+                //Don't create an invoice for project past its end date, and just mark it as AwaitingCloseout
+                if (project.End < now)
+                {
+                    project.Status = Project.Statuses.AwaitingCloseout;
+                    await _dbContext.SaveChangesAsync();
```
jSylvestre, line 74: Should we generate an email at this time to the FM that this project is awaiting closeout?

jSylvestre (approved):
Other than a few of my comments. Seems ok and better than before.

Harvest.Core/Extensions/DateTimeExtensions.cs
sprucely: It's just shorthand for if/else-if/else-if/else

Harvest.Core/Services/InvoiceService.cs
sprucely: Maybe? But I think that should be a separate issue.

PR #450, opened 2021-08-19
Replaced Input for Typeahead for Expenses

### LINKED ISSUE #447: Change dropdown lists for expenses/rates to be typeahead lists
Some of the lists can get pretty long.  It shouldn't be too hard to use the typeahead to allow searching.  Everything else, like validation & calculations, should continue to work the same way.

ex: https://localhost:5001/expense/entry/38

wongband:
resolve #447

Todo:
Scott's second change request


Harvest.Web/ClientApp/src/App.tsx
```diff
@@ -22,6 +22,7 @@ import { InvoiceListContainer } from "./Invoices/InvoiceListContainer";
 import { TicketCreate } from "./Tickets/TicketCreate";
 import { TicketsContainer } from "./Tickets/TicketsContainer";
 import { TicketDetailContainer } from "./Tickets/TicketDetailContainer";
+import 'react-bootstrap-typeahead/css/Typeahead.css';
```
srkirkland, line 25: why do we need this now where we didn't before?  It should already be in the index.tsx page and thus doesn't need to be declared twice.

srkirkland (changes requested):
Two things I'd like to see changed, not exactly sure how to do the second. 
1. I'd like to have an easy way to clear out an option once selected, like they do in http://ericgio.github.io/react-bootstrap-typeahead/#controlling-selections.  Once it is cleared out, we need to reset the state as if nothing had been selected.  This is happening in the main branch with dropdowns but not here.
2. The options should be restricted to just what we offer.  It would be nice if people couldn't type random stuff into the box.

Harvest.Web/ClientApp/src/App.tsx
wongband: Sorry didn't notice that. I will get rid of it.

jSylvestre (changes requested):
On the quote page, this is not triggering the enable of the submit button

jSylvestre (changes requested):
Save a quote, then open it back up and the info isn't populated:
image

jSylvestre (changes requested):
Almost there.
The dropdown should extend out of the container:

image

jSylvestre (approved):
Works for me, but maybe wait for @srkirkland to take a look.

srkirkland (approved):

PR #452, opened 2021-08-23
yup type error messages

jSylvestre:


sprucely (approved):

Harvest.Web/ClientApp/src/schemas.ts
```diff
@@ -46,12 +46,12 @@ export const workItemSchema: SchemaOf<WorkItem> = yup.object().shape({
   activityId: yup.number().required().integer(),
   type: yup.mixed().required().oneOf(["Acreage", "Equipment", "Other", "Labor"]),
   rateId: yup.number().required().default(0),
-  rate: yup.number().required().positive("Work item rate must be a positive number"),
+  rate: yup.number().required().typeError("Work item rate must be a positive number").positive("Work item rate must be a positive number"),
```
srkirkland, line 49: @jSylvestre I hate to jump in and be that guy but let's not duplicate the same exact string so many times.  I can let it go if you just want to move on with this PR but otherwise it seems like moving this repeated string to a simple `const` variable or in a separate ErrorMessages.ts file would be nice.

srkirkland (changes requested):

PR #453, opened 2021-08-23
Duplicate errorMessages in new file...

jSylvestre:


sprucely (approved):

sprucely:
At some point we may want to move all error messages to errorMessages.ts, but I think this is good enough for now.

srkirkland (approved):
ahh nice!

PR #454, opened 2021-08-23
update precision of some decimals

srkirkland:
also change the way we create approved expenses

### DIFF
--- a/Harvest.Core/Domain/Expense.cs
+++ b/Harvest.Core/Domain/Expense.cs
@@ -20,7 +20,7 @@ namespace Harvest.Core.Domain
         [StringLength(250)]
         public string Description { get; set; }
 
-        [Display(Name = "Rate")] 
+        [Display(Name = "Rate")]
         [Range(0.01, Double.MaxValue, ErrorMessage = "Rate must be greater than 0.01")]
         public decimal Price { get; set; }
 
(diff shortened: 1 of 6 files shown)


sprucely (approved):

Harvest.Core/Domain/Expense.cs
```diff
@@ -62,9 +62,9 @@ internal static void OnModelCreating(ModelBuilder modelBuilder)
         {
             modelBuilder.Entity<Expense>().HasIndex(a => a.ProjectId);
 
+            modelBuilder.Entity<Expense>().Property(a => a.Price).HasPrecision(18, 2);
             modelBuilder.Entity<Expense>().Property(a => a.Total).HasPrecision(18, 2);
-
-            modelBuilder.Entity<Expense>().Property(a => a.Approved).HasDefaultValue(true);
+            modelBuilder.Entity<Expense>().Property(a => a.Quantity).HasPrecision(18, 2);
```
jSylvestre, line 67: I'm wondering about quantity. This doesn't have to be hours, it could be something where we want more precision

srkirkland: @jSylvestre I think it's worth debating, I'm not sure the correct answer.  In the slack channel we just picked this since we use it for many other things.  However if we are going to go with 2 decimal places only we should probably enforce that client-side w/ validation too....

PR #458, opened 2021-08-24
Job to send out closeout notifications

### LINKED ISSUE #429: Project closeout
Supervisor/FM roles should have functionality to "closeout" a project.  This will show unbilled expenses as well as how much is left in the budget.  Once the user is satisfied, they will click a button which creates the "final" invoice and moves the project to a Completed state (or at least a state where no more expenses can be entered).

jSylvestre:
resolve #429


srkirkland (changes requested):
Hopefully this isn't too much work but please change the job to just be `Harvest.Jobs.Notifications` because there is a decent change there will be more notifications and it'd be better to have one job process these than a job per type of notification.

sprucely (approved):
LGTM. I honestly don't recall the reasoning on choice of service lifetimes. Wondering if we should just make everything singleton for jobs.

jSylvestre:
@srkirkland This would have to be merged to fix those errors, correct?

PR #463, opened 2021-08-27
Project Closeout

sprucely:
Need to `npm install`

I extended usePromiseNotification to accept callbacks that can customize the success and error messages based on the value returned.

I also replaced a few modals with a useConfirmationDialog hook that reduces a bunch of boilerplate and allows awaiting a getConfirmation(). This one has been bugging me for some time, and I think the result is much more reasonaboutable...
```
  const [getConfirmation] = useConfirmationDialog({
    title: "Closeout Project",
    message: (
      <ul>
        <li>Generates a final invoice if there are any unbilled expenses</li>
        <li>Sets project status to either Completed or CloseoutPending based on whether there are any pending invoices</li>
      </ul>)
  });
```
And later used...
```
  const closeoutProject = async () => {
    if (!await getConfirmation()) {
      return;
    }
    ...
```


Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+interface RouteParams {
+  projectId?: string;
+}
+
+interface FinalAcreageExpense {
+  amount: number;
+}
+
+const finalAcreageExpenseSchema = yup.object().shape({
```
jSylvestre, line 28: Shouldn't this be in the schema file?

jSylvestre:
Is there a closeout button somewhere?

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+      <div className="card-content card-green-bg">
+        <div className="row">
+          <div className="col text-right">
+            <Button id="CloseoutButton" color="primary" disabled={!canCloseout || notification.pending} onClick={closeoutProject}>
+              Closeout Project <FontAwesomeIcon icon={faCheck} />
+            </Button>
+            <ShowFor condition={!canCloseout}>
+              {/* Not sure if there's a better way to handle indicators as to why something is disabled */}
+              <UncontrolledTooltip target="CloseoutButton" >
+                Project is not awaiting closeout, and it doesn't have an end date that falls within the last seven days or later.
```
jSylvestre, line 178: For my emails, I just picked 7 days.
I don't think we want to limit the closeout to that.

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
sprucely: I figured it was small schema that would not be referenced anywhere else. But I can move it.

sprucely:
> 
> 
> Is there a closeout button somewhere? (To get to the closeout container)

https://github.com/ucdavis/Harvest/blob/cde393dba67885a9be6c12c7c6624ad37ec7921b/Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx#L86-L99

Harvest.Core/Domain/Project.cs
```diff
@@ -175,6 +175,7 @@ public class Statuses
             public const string ChangeRequested = "ChangeRequested";
             public const string Completed = "Completed";
             public const string AwaitingCloseout = "AwaitingCloseout";
+            public const string FinalInvoicePending = "FinalInvoicePending";
 
             public static List<string> TypeList = new List<string>
             {
```
srkirkland, line 181: Need to add your new `FinalInvoicePending` status to this list.  Also, please order them more or less in their active order, so completed should be last with the awaiting closeout and final invoice ones just before.

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
srkirkland: yea might as well for consistency

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+    formErrorCount,
+  } = useInputValidator<FinalAcreageExpense>(finalAcreageExpenseSchema);
+
+  useEffect(() => {
+    const cb = async () => {
+      const response = await fetch(`/Project/Get/${projectId}`);
+      if (response.ok) {
+        const proj: Project = await response.json();
+        setProject(proj);
+        setBeyondCloseoutDisplayDate(new Date().getTime() - new Date(proj.end).getTime() >= (7 * millisecondsPerDay));
```
srkirkland, line 58: ugh date logic in JS is horrible.  but this is a pretty good way to do it.

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+      return;
+    }
+
+    const request = fetch(`/Invoice/DoCloseout/${projectId}`, {
+      method: "POST"
+    });
+
+    setNotification(request, "Closing Out Project", async (response: Response) => {
+      const result = await response.json() as Result<number>;
+      setCloseoutRequested(true);
```
srkirkland, line 112: too bad you can't just redirect here and have to go through the useEffect dance with the extra setState variable.

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+      return result.message;
+    });
+  }
+
+  if (project === undefined) {
+    return <div>Loading ...</div>;
+  }
+
+  const canCloseout = project.status === "AwaitingCloseout" ||
```
srkirkland, line 121: How does a project get this status in the first place?

Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
```diff
@@ -29,32 +40,36 @@ export const UnbilledExpensesContainer = () => {
...
 
     cb();
-  }, [projectId]);
+  }, [projectId, props.newExpenseCount]);
 
-  // This function closes the modal and deletes the entry from the db
-  const confirmModal = async (expenseId: number) => {
-    const response = await fetch(`/Expense/Delete?expenseId=${expenseId}`, {
+  const deleteExpense = async (expense: Expense) => {
+    if (! await confirmRemoveExpense()) {
```
srkirkland, line 51: make sure to run prettier if you haven't, the space here after the bang strikes me as unformatted.

Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
```diff
@@ -25,14 +25,19 @@ export const ProjectDetailContainer = () => {
...
     const cb = async () => {
       const response = await fetch(`/Project/Get/${projectId}`);
 
       if (response.ok) {
-        setProject(await response.json());
+        const project = await response.json() as Project;
+        setProject(project);
+
+        //show Closeout link if we are in last week of or beyond project end 
+        setBeyondCloseoutDisplayDate(new Date().getTime() - new Date(project.end).getTime() >= (7 * millisecondsPerDay));
```
srkirkland, line 40: This calculation seems handy enough to use a few times and probably will come in handy again.  Perhaps instead you can move this to the Calculation Utils itself as a function like `getDaysDiff(date1, date2)` and then here it would be easier to read that you are just checking to see if the date diff is >= 7.

srkirkland (commented):
Code looks good to me.  I don't particularly like how the modals are now hidden behind another npm lib and some `useRef` complexity but I don't have a big problem with it either.  Invoice logic looks good from what I can tell.

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
sprucely: I don't recall if I actually tried that and had an issue with it, or if it just looked like it would cause an issue. `¯\_(ツ)_/¯`

Harvest.Web/ClientApp/src/Closeout/CloseoutContainer.tsx
sprucely: This is set automatically by the invoice service when the date is past project end.

jSylvestre (approved):
I think Scott had some changes, but looks ok to me.
May need Cal to style the extra closeout button (when there are too many buttons, they wrap onto another line)

PR #464, opened 2021-08-31
Show System if you have System

jSylvestre:


Harvest.Web/Views/Permissions/Delete.cshtml
```diff
@@ -23,16 +23,26 @@
...
+                          <br/>
+                      }
+                  }else{
+                      foreach (var permission in Model.Permissions.Where(a => a.Role.Name != Role.Codes.System))
+                      {
+                          <input id="@permission.Role.Id" type="checkbox" value="@permission.Role.Id" name="roles"/>
+                          <label for="@permission.Role.Id">@permission.Role.Name</label>
+                          <br/>
+                          <br/>
+                      }
```
sprucely, line 43: I would have done `.Where(a => ViewBag.ShowSystem || a.Role.Name != Role.Codes.System)` in order to avoid duplicate blocks.

sprucely (approved):
Mostly what I would have done. 😃

Harvest.Web/Controllers/PermissionsController.cs
```diff
@@ -52,20 +61,33 @@ public async Task<IActionResult> Index()
 
         public async Task<IActionResult> Create()
         {
+            IQueryable<Role> rolesQuery = _dbContext.Roles;
+
+            //If you have System, show system.
+            if (!await _userService.HasAccess(AccessCodes.SystemAccess))
+            {
+                rolesQuery = rolesQuery.Where(a => a.Name != Role.Codes.System);
+            }
```
srkirkland, line 70: This code is duplicated several times but otherwise PR looks good

Harvest.Web/Views/Permissions/Delete.cshtml
jSylvestre: I changed it to filter server side

PR #465, opened 2021-08-31
Enable nightly invoices

sprucely:
The new setting will only create invoices if there are unbilled expenses.


Harvest.Web/Startup.cs
```diff
@@ -140,7 +140,8 @@ public void ConfigureServices(IServiceCollection services)
             services.Configure<SparkpostSettings>(Configuration.GetSection("SparkPost"));
             services.Configure<StorageSettings>(Configuration.GetSection("Storage"));
             services.Configure<EmailSettings>(Configuration.GetSection("Email"));
-            
+            services.Configure<DevSettings>(Configuration.GetSection("Dev"));
```
jSylvestre, line 143: Do you need this here? Isn't this just a job setting?

jSylvestre: Never mind. I see it now.

jSylvestre (approved):
Didn't test it, but seems fine to me.

srkirkland (approved):

PR #466, opened 2021-08-31
typo

jSylvestre:


srkirkland (approved):
probably should just be `_` since that's the designator for a variable in an arrow function that won't be used.  But that's just nitpicking this is fine and much better than before of course

PR #468, opened 2021-09-01
Ensure mounted before setting state after fetch/async ops

sprucely:
My idea about copy/pasting a useEffect snippet was flawed. React gave a warning about needing to store the isMounted variable in a ref in order to not get erased on subsequent render. So I put it in a useIsMounted hook after all. For consistency I used the hook everywhere rather than the hook some places and
 ```
let isMounted=true; 
doAsyncStuff(); 
return () => {isMounted = false};
```
elsewhere.

I'm not married to this. I could just do the above in async effects and use the hook in the async functions.


srkirkland (approved):
I like the concept and execution

PR #502, opened 2021-09-03
Added a modal to display large requirements

### LINKED ISSUE #499: Big Requirements?
If the requirements have a lot of text it takes up a lot of space everywhere.  If it's longer than 256 characters, truncate with an `...` ellipsis and add a `View More` button.  This button when clicked would open a large modal which shows all of the requirements text, even if it's ver long.

image

wongband:
resolve #499


Harvest.Web/ClientApp/src/Shared/ProjectHeader.tsx
```diff
@@ -7,6 +9,25 @@ interface Props {
 
 export const ProjectHeader = (props: Props) => {
   const { project, title } = props;
+  const [modal, setModal] = useState<boolean>(false);
+  const toggleModal = () => setModal(!modal);
```
srkirkland, line 13: whenever you set state based on existing state, you should use the callback to ensure you are working with the proper value.  So here it should be `const toggleModal = () => setModal(m => !m);`

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Shared/ProjectHeader.tsx
```diff
@@ -7,6 +9,25 @@ interface Props {
...
+
+  // Function to show a truncated version of project.requirements if the
+  // character count is above 256
+  const projectReqs = (requirements: string) => {
```
srkirkland, line 17: Also here, this is pretty minor but you are recalculating this every time the component runs although the requirements prop isn't going to change.  It'd be a good place to `useMemo`

jSylvestre:
Should the model have a title or anything?

image

srkirkland (approved):
Good changes -- I agree with Jason, just add a modal header saying "Project Requirements" or similar, and then it'll be 💯

PR #503, opened 2021-09-03
Remove ability for supervisor to submit quotes

### LINKED ISSUE #469: Supervisor should not be able to submit quote (save only)

srkirkland:
Removed from front end and added backend permissions check.

Closes #469

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
@@ -20,6 +20,7 @@ import { useInputValidator, ValidationProvider } from "../FormValidation";
 import { quoteContentSchema } from "../schemas";
 import { checkValidity } from "../Util/ValidationHelpers";
 import { useIsMounted } from "../Shared/UseIsMounted";
+import { ShowFor } from "../Shared/ShowFor";
 
 interface RouteParams {
   projectId?: string;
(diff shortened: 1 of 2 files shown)


Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
```diff
@@ -285,22 +286,34 @@ export const QuoteContainer = () => {
...
+              <ShowFor roles={["Supervisor"]}>
+                <button
+                  className="btn btn-primary mt-4"
+                  onClick={() => save(false)}
+                  disabled={notification.pending || formErrorCount > 0}
+                >
+                  Save Quote
+                </button>
+              </ShowFor>
+              <ShowFor roles={["FieldManager"]}>
```
jSylvestre, line 299: Didn't test, but I think you need bother roles here?

sprucely: I don't think so, because the prior `<ShowFor/>` handles `Supervisor` role.

jSylvestre (approved):

Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
srkirkland: yea, I wanted to get fancy here and have different buttons be styled different depending on the role.  This seemed the cleanest way to do it.

PR #504, opened 2021-09-03
don't overwrite original project with new requirement changes

### LINKED ISSUE #481: Change Request form project details need to be duplicated
When change fields in the change request, any changes to fields also changes the header info, which should never change

srkirkland:
Created a new state variable to track the original project, which only gets set on initial api load.

closes #481


jSylvestre (approved):
🚢

PR #508, opened 2021-09-07
Humanized crops and project status

### LINKED ISSUE #484: Tags should be humanized CamelCase -> Camel Case
Any where they are shown, especially proj details and project list.  Can probably use something like this https://stackoverflow.com/questions/4149276/how-to-convert-camelcase-to-camel-case.  If that works, or whatever does instead, put it in a util file, perhaps StringFormatting.ts or similar, and then import to use.

### LINKED ISSUE #486: humanize crop tags
At least add a nice space between the list of crops (corn,celery -> "corn, celery").  Just do a split() on comma, and then join() back together with a command and space.

wongband:
resolve #486 
resolve #484


Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
```diff
@@ -73,7 +73,13 @@ export const ProjectTable = (props: Props) => {
...
         Cell: (data: Cell<Project>) => (
           <span className={`project-status-${data.row.original.status}`}>
-            {data.row.original.status}
+            {data.row.original.status
+              // Regex to insert a space before all caps
+              .replace(/([A-Z])/g, " $1")
+              // Regex to uppercase the first word
+              .replace(/^./, (str) => {
+                return str.toUpperCase();
+              })}
```
srkirkland, line 82: Move this into a helper somewhere in the utils folder.  It's non-intuitive enough and it needs to be used >1 so it'll be better there.

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
wongband: Gotcha, wasn't sure if we needed to do that

jSylvestre (approved):

Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
```diff
@@ -7,6 +7,7 @@ import { ReactTable } from "../Shared/ReactTable";
 import { ReactTableUtil } from "../Shared/TableUtil";
 import { Project } from "../types";
 import { formatCurrency } from "../Util/NumberFormatting";
+import { convertCamelCase } from "../Util/ConvertCamelCase";
```
srkirkland, line 10: I'm gonna nitpick a little here -- you have a non-default export called `convertCamelCase` from a file called `ConvertCamelCase`.  What other functions are you expecting to export?  I think you have two good choices here -- 1) would be change the file name to something more generic, like `StringFormatting.ts`, or (b) make the export default, since the file only does one thing (but then if you add more you'll have to change it).

srkirkland (changes requested):

Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
wongband: I will go with choice 1 just in case we need other string manipulation functions later on.

srkirkland (approved):

PR #512, opened 2021-09-07
prevent FM and Supervisor roles

jSylvestre:


srkirkland (approved):
I meant to do the IF/ELSE in react for the buttons.  But I also think this is fine too, at least for now.  Probably wouldn't hurt to be specific about the roles.

PR #513, opened 2021-09-07
Fixed some small issues

### LINKED ISSUE #470: remove lorem ipsum
Remove lorem ipsum and other fake text from the app.  There is some on the homepage, not sure about anywhere else.

### LINKED ISSUE #477: rename Approve Quote button
For now let's go with `View Quote` instead.   This is for projects in the pending approval state.

### LINKED ISSUE #483: Change notification text of QuoteContainer based on submit vs. save
in quote container, currently the notification always says "Saving Quote".  If the `submit` bool is true, instead it should be `Submitting Quote`.

wongband:
resolve #470 
resolve #477 
resolve #483

### DIFF
--- a/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
@@ -179,7 +179,7 @@ export const QuoteContainer = () => {
       body: JSON.stringify(quote),
     });
 
-    setNotification(request, "Saving Quote", "Quote Saved");
+    setNotification(request, "Submitting Quote", "Quote Saved");
 
     // TODO: add progress and hide info while saving
     const saveResponse = await request;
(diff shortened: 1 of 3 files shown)


Harvest.Web/ClientApp/src/Quotes/QuoteContainer.tsx
```diff
@@ -179,7 +179,7 @@ export const QuoteContainer = () => {
       body: JSON.stringify(quote),
     });
 
-    setNotification(request, "Saving Quote", "Quote Saved");
+    setNotification(request, "Submitting Quote", "Quote Saved");
```
srkirkland, line 182: This should say "Quote Submitted" when finished just for `submitted` status

jSylvestre, line 182: You need to check if it is submitting or saving.
Then change the "Saved" part as well

srkirkland (changes requested):

jSylvestre (approved):
Didn't test, but that looks correct to me.

srkirkland (approved):

PR #514, opened 2021-09-07
Dynamic Code Splitting because of the PDF library

srkirkland:
The react-pdf library is adding 812KB to our JS, while our entire app, including vendor JS, is otherwise 318KB.  Additionally, the pdf lib gives an ASM.js warning in the console which is annoying.

So, my solution is to use code-splitting to move the PDF components into dynamically loaded components.  There are quite a few ways to do this, but using `React.lazy()` seemed like an easy plan. See https://reactjs.org/docs/code-splitting.html.

This works really well, and this all made me also realize there was no reason for me to include all chunks on the rendered layout page by default.  I can just include the runtime, main, and primary chunk (determined by whatever `npm run build` spits out) and everything will work, it'll only load what is needed, and it'll dynamically load other stuff later.

### DIFF
--- a/Harvest.Web/ClientApp/src/Invoices/InvoiceDetailContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Invoices/InvoiceDetailContainer.tsx
@@ -1,15 +1,13 @@
-import { useEffect, useState } from "react";
+import React, { Suspense, useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
-import { PDFDownloadLink } from "@react-pdf/renderer";
 
 import { ProjectWithInvoice } from "../types";
-import { InvoicePDF } from "../Pdf/InvoicePDF";
 import { ProjectHeader } from "../Shared/ProjectHeader";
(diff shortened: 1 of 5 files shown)


sprucely (approved):
LGTM. So what's special about `3.[hash].chunk.js` that it needs to be included but not `2.[hash].chunk.js`

srkirkland:
> LGTM. So what's special about `3.[hash].chunk.js` that it needs to be included but not `2.[hash].chunk.js`?

No idea! But when you do the prod build the generated html file from CRA only includes 3.* and not the others. So if you were running a normal react app that's all that is on the html. But I do not know what their logic for that is.

jSylvestre (approved):
I only tested the quote PDF, looks ok to me.
I did get a map error in the console but I think that was there before.

jSylvestre:
@srkirkland Do we need an error boundary?

srkirkland:
> @srkirkland Do we need an error boundary?

Probably don't need one, since it's not much different than it was before in terms of error likelyhood.  But could it be a fun experiment, yea maybe.

PR #515, opened 2021-09-08
Sort, filter, and link with the whole field

jSylvestre:
image


srkirkland (approved):
looks good but is this prettier formatted?  seems like a really long line?  if it is, then a-ok

sprucely (approved):
Works for me.

PR #518, opened 2021-09-08
Massive NPM Package Upgrade

srkirkland:
NOTE: You will want to delete your whole node_modules folder before doing a `npm install` because of a change to development-only browser-list.

Recreated the `package-lock.json` file in the current npm format, and rebuilt to remove lots of security warnings from older packages deep in the dependency tree.

Updated all packages except for bootstrap (since we want to stay on v4).  Minor modifications made to code to handle changes to leaflet and the datepicker types.

New prettier/eslint rules will require minor formatting changes to a few files, but I will make those in a future PR to keep this clean.


sprucely (approved):
Worked for me. No errors. Npm audit showed 3 moderate vulnerabilities in browserslist and glob-parent.

jSylvestre (approved):

PR #519, opened 2021-09-08
Don't allow accounts with 0% to be submitted

### LINKED ISSUE #479: Validation on account change/approval -- don't allow account with 0%
Show an error and don't allow save/submit if there are any 0% accounts.  We already check that the total percentage has to be 100% so this would probably be pretty simple to just add the extra logic wherever that is.

wongband:
resolve #479


Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
```diff
@@ -10,6 +10,7 @@ import { ProjectAccount } from "../types";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faMinusCircle } from "@fortawesome/free-solid-svg-icons";
 import { useIsMounted } from "../Shared/UseIsMounted";
+import { array } from "yup";
```
sprucely, line 13: Don't see where this is being used.

sprucely (approved):

jSylvestre (changes requested):
Still seeing this after I delete the row:

image

Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
wongband: It got added when I typed in array on accident. I will get rid of it

jSylvestre:
That might be happening because if I delete an account, then add it back it remembers the %

jSylvestre:
I don't know if it is worthwhile to fix this, but if you add 2 accounts so they both add up to 100%, then remove one you don't see an error message but the Approve Quote button is still disabled, so very minor:

image
image

jSylvestre (approved):

PR #521, opened 2021-09-08
Added alert for project change requests

### LINKED ISSUE #482: project details change requests
add a big alert to the top of a change request which says which project it is a change request for (with link back to original project).  Use a regular bootstrap alert div `<div class="alert alert-info">`

wongband:
resolve #482

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
@@ -149,12 +149,14 @@ export const RequestContainer = () => {
 
   return (
     <div>
-      <div className="alert alert-info">
-        You are making a change request for {project.name}{" "}
-        <Link className="alert-link" to={`/project/details/${project.id}`}>
-          Click here to go back to the project details page
-        </Link>
-      </div>
+      {projectId && (
+        <div className="alert alert-info">
+          You are making a change request for {project.name}{" "}
+          <Link className="alert-link" to={`/project/details/${project.id}`}>
+            Click here to go back to the project details page
+          </Link>
+        </div>
+      )}
       <div className={originalProject && "card-wrapper"}>
         {originalProject !== undefined && (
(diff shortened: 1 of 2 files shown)


Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
```diff
@@ -147,166 +147,176 @@ export const RequestContainer = () => {
...
-                        clearIcon={null}
-                        value={project.start}
-                        onChange={(date: Date) =>
-                          setProject({ ...project, start: date })
-                        }
-                      />
+    <div>
+      <div className="alert alert-info">
+        You are making a change request for {project.name}{" "}
+        <a href={`/project/details/${project.id}`} className="alert-link">
```
srkirkland, line 153: @wongband We don't ever want to use `<a>` inside a react app since that doesn't cause the client-side react-router routing.

srkirkland (changes requested):

srkirkland (changes requested):
the already should only show up when it is actually a change request.  As it is now, it shows up during any new request.

Harvest.Web/ClientApp/src/Requests/RequestContainer.tsx
```diff
@@ -147,166 +148,178 @@ export const RequestContainer = () => {
...
 
   return (
-    <div className={originalProject && "card-wrapper"}>
-      {originalProject !== undefined && (
-        <ProjectHeader
-          project={originalProject}
-          title={"Original Field Request #" + (originalProject?.id || "")}
-        />
+    <div>
+      {projectId && (
```
srkirkland, line 152: @wongband This works but to be consistent you should use `originalProject !== undefined` as your check

srkirkland (changes requested):

jSylvestre (approved):

srkirkland (approved):

PR #522, opened 2021-09-09
Jcs/hacky date sort filter fix

### LINKED ISSUE #517: Make filtering work for date columns
use custom filter function or build a custom filter, or possibly just change the accessor but we'll need to make sure sorting still works correctly in all cases.  if we want to get fancy we can make a custom filter of a datepicker which can select a range or single date and we filter to those.

jSylvestre:
Resolve #517 
This works, because the accessor data looks like this:
2021-12-04T08:00:00 12/4/2021

so sorting is always correct, and if you filter you can use 12-04 or 12/4 etc.


srkirkland (approved):
Ok I'm fine with this approach. Can you check before you merge to make sure there aren't other instances of dates in our tables? Invoice or tickets or something maybe or expenses?

PR #526, opened 2021-09-09
Have an endpoint for the invoice list

### LINKED ISSUE #525: Invoice list route
https://harvest-test.azurewebsites.net/project/invoices/2

jSylvestre:
Resolve #525


srkirkland (commented):
Can I suggested alternately we move these endpoints to the `InvoiceController`?  At least the one for Listing Invoices.  That would have the benefit of keeping the routing too.

jSylvestre:
@srkirkland I don't know why, but I try to move stuff into the invoice controller and the projectId is always 0

srkirkland (approved):

PR #527, opened 2021-09-09
Cyd/sitecleanup 4

jSylvestre:


Harvest.Web/ClientApp/src/sass/_layout.scss
```diff
@@ -18,7 +18,6 @@ body {
   border-right: 1px solid $borders;
   border-bottom: 1px solid $borders;
   box-shadow: $box-shadow;
-  overflow: hidden;
```
sprucely, line 21: Does this mean we no longer have to manually put `overflow: "visible"` on card-wrapper divs that have DateTime and/or Typeahead components?

sprucely (approved):

Harvest.Web/ClientApp/src/sass/_layout.scss
cydoval: yea

PR #529, opened 2021-09-10
Cancel request

### LINKED ISSUE #485: Projects/Change Requests that will not be approved?
What if there is a project that will never get a quote or be approved?

What if we have a change request that needs to be deleted?

jSylvestre:
Resolve #485 

Probably a good idea to change this so it has a confirmation where they have to type Yes.
Make it look like the reject quote?


srkirkland (approved):
Adds to our button soup but the implementation looks good. I agree it definitely needs a confirm though. And I'd say it needs a history but probably no email. I'm wondering though should we set active to false for these? Or just add filters to everywhere else? I'd lean making it inactive i think.

PR #534, opened 2021-09-10
Validate All

sprucely:
`<InputErrorMessage />` now displays multiple errors, each in its own `<p/>`.
Added a `validateAll`, which will validate only inputs that have a corresponding `<InputErrorMessage />`.

Wasn't sure about storing validate callbacks in a ref. I guess I could have stored in context state a validateAllCount/setValidateAllCount, increment it each time `validateAll` is called, and use an effect on that to trigger validations in each hook instance. Doesn't seem much prettier than refs, but I can do it if you think it would be easier to follow.

I don't know why the RequestContainer diff is so large. I didn't change THAT much. Maybe to do with resolving conflicts while rebasing?

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/SearchPerson.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/SearchPerson.tsx
@@ -1,18 +1,22 @@
 import React, { useState } from "react";
 
-import { AsyncTypeahead, Highlighter } from "react-bootstrap-typeahead";
+import {
+  AsyncTypeahead,
+  Highlighter,
+  TypeaheadProps,
+} from "react-bootstrap-typeahead";
 import { useIsMounted } from "../Shared/UseIsMounted";
 
 import { User } from "../types";
 
-interface Props {
+interface Props extends Pick<TypeaheadProps<string>, "onBlur"> {
   user?: User;
-  setUser: (user: User) => void;
+  onChange: (user: User | undefined) => void;
 }
 
 export const SearchPerson = (props: Props) => {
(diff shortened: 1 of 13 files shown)


jSylvestre:
Don't know if it is specific to this PR, but the validation of off on this quote:
image

Harvest.Web/ClientApp/src/Requests/SearchPerson.tsx
```diff
@@ -1,18 +1,22 @@
...
+interface Props extends Pick<TypeaheadProps<string>, "onBlur"> {
   user?: User;
-  setUser: (user: User) => void;
+  onChange: (user: User | undefined) => void;
 }
 
 export const SearchPerson = (props: Props) => {
   const [isSearchLoading, setIsSearchLoading] = useState<boolean>(false);
-  const [users, setUsers] = useState<User[]>([]);
+  const [users, onChanges] = useState<User[]>([]);
```
sprucely, line 19: oops. I'll fix this momentarily...

jSylvestre (approved):
Little different behavior with save, but I think it is better this way.

srkirkland (approved):
Looks good to me.  Didn't really review requestcontainer because of all the changes but the rest looked fine.  The validation stuff getting this complex in the background is a bit scary, especially with all the react-internals stuff, so keeping that current will be a challenge.  But the end-dev and user experience is pretty nice so overall 👍 .  I think we might want to consider breaking out the validation library kinda stuff, like useInputValidator and your ValidationProvider into it's own NuGet package that we can cover with tests and then utilize in different projects.

Issue #523, opened 2021-09-09
Crops?


jSylvestre:
Do we want a table of crops and a typeahead to search for them?
As well as the admin side to add/remove the list?

srkirkland:
@jSylvestre I kinda think we should

jSylvestre:
Populate the react info from the DB...

PR #535, opened 2021-09-13
Jcs/crop lookup

jSylvestre:
This has the DB part of the issue if someone else wants to do the typeahead change
Part of #523


Harvest.Web/Controllers/Api/LookupController.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+
+        public async Task<ActionResult> SearchCrops(string type, string query)
+        {
+            //I think this is right, didn't test
+            var crops = await _dbContext.CropLookups
+                .Where(a => a.Type == type &&
+                            (EF.Functions.Like(a.Crop, query.EfStartsWith()) ||
+                             EF.Functions.Like(a.Crop, query.EfContains())))
```
sprucely, line 28: I think `EfStartsWith()` is a subset of `EfContains()` and so is redundant here.

sprucely (approved):
Looks good. Maybe just remove that unnecessary `EfStartsWith`

srkirkland (changes requested):
first is I hate to have you redo the migrations but I think the classname here should be `Crop` and the table thus `Crops`.  We do lookups all the time in our DBs but don't usually use the `Lookups` naming scheme and I'm not a fan of that specificity. 

second is more simple and doesn't really need to be fixed, but I think it's unlikely we'll ever bother searching for crops.  A simple controller method to just return all crops would probably be just fine, the list will not be that long, and it'll make searching much easier via typeahead.

Harvest.Web/Controllers/Api/LookupController.cs
jSylvestre: @sprucely I based it off what we do in Peaks

srkirkland (approved):

PR #537, opened 2021-09-13
Jcs/crop lookup crud

jSylvestre:


Harvest.Web/Startup.cs
```diff
@@ -209,7 +209,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, AppDbCon
                     name: "default",
                     pattern: "{controller}/{action}/{id?}",
                     defaults: new { controller = "Home", action = "Index" },
-                    constraints: new { controller = "(invoice|rate|permissions)" }
+                    constraints: new { controller = "(invoice|rate|permissions|crop)" }
```
jSylvestre, line 212: This was fun. Without this, I had to put the id as a hidden value on the edit form, otherwise it returned zero in the controller id post.

sprucely (approved):

Harvest.Web/Startup.cs
jSylvestre: @srkirkland Also, should I remove invoice from here?

srkirkland: yea

srkirkland (approved):

PR #541, opened 2021-09-14
Added some fancy filtering

### LINKED ISSUE #531: Fancy table filtering
A few other fancy table filters, these will be easier than the date filter one:

- [x] Crop Type filter should just be a dropdown with "Any, Row, Tree" as the options
- [x] Progress filter should be dropdown with "<25%, 25-50%, 50-75%, >75%" as the options.
- [x] Status should be a dropdown with all of the possible statuses, or at least all of the statuses that are in the table

wongband:
resolve #531

### DIFF
--- a/Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
@@ -5,7 +5,11 @@ import { Progress } from "reactstrap";
 
 import { ReactTable } from "../Shared/ReactTable";
 import { ReactTableUtil } from "../Shared/TableUtil";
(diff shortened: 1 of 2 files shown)


jSylvestre:
@wongband  I don't know if you can fix or if this would be a @cydoval thing, but these get chopped once you select longer text

image

jSylvestre (changes requested):
This should be about 60%:
image

But I don't see it when I do <75%
image

wongband:
@jSylvestre Sorry about that, it should be ">75". I will fix it right now

jSylvestre (approved):

srkirkland (approved):
code looks good

PR #544, opened 2021-09-15
Get crops from DB

sprucely:
I was originally going to have the typeahead load crops as needed, but ended up just loading them all ahead of time based on crop type. So CropController/Search has an optional query parameter that isn't being used by the client. I don't think it hurts to leave it there.

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/Crops.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/Crops.tsx
@@ -58,7 +58,7 @@ export const Crops = (props: Props) => {
   };
 
   return (
-    <Typeahead<string>
+    <Typeahead
       id="crops" // for accessibility
       allowNew
       multiple
--- a/Harvest.Web/ClientApp/src/Requests/Crops.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/Crops.tsx
@@ -30,25 +30,18 @@ export const Crops = (props: Props) => {
     onSearch();
   }, [props.cropType]);
 
-  const onSelect = (selected: (string | { label: string })[]) => {
+  const onSelect = (selected: string[]) => {
     if (selected && selected.length > 0) {
-      // selections are either strings or an object with the string in a label prop for custom fields, so project them to just the strings we want
-      const selectedStrings = selected.map((s) =>
-        typeof s === "string" ? s : s.label
-      );
-
-      props.onChange(selectedStrings.join(","));
+      props.onChange(selected.join(","));
     } else {
       props.onChange(""); //If it is cleared out...
     }
   };
 
-  const onSearch = async (query?: string) => {
+  const onSearch = async () => {
     setIsSearchLoading(true);
 
-    const response = await fetch(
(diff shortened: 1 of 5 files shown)


Harvest.Web/ClientApp/src/Requests/Crops.tsx
```diff
@@ -40,8 +43,29 @@ export const Crops = (props: Props) => {
...
+
+        getIsMounted() && setOptions(crops);
+      }
+    }
+    getIsMounted() && setIsSearchLoading(false);
+  };
+
   return (
-    <Typeahead
+    <Typeahead<string>
```
srkirkland, line 68: is this newly required?  never seen this syntax inside component before, looks different

Harvest.Web/ClientApp/src/Requests/Crops.tsx
```diff
@@ -40,8 +43,29 @@ export const Crops = (props: Props) => {
     }
   };
 
+  const onSearch = async (query?: string) => {
+    setIsSearchLoading(true);
+
+    const response = await fetch(
+      `/Crop/Search?type=${props.cropType}${(query && `&query=${query}`) || ""}`
```
srkirkland, line 50: I'm ok with leaving the query param in the server method but I would rather you just remove it from here to make it more readable.

srkirkland:
Also seems like this broke the tests

Harvest.Web/Controllers/CropController.cs
```diff
@@ -142,5 +142,17 @@ public async Task<ActionResult> Delete(int id, Crop model)
                 return View();
             }
         }
+
+        // GET: CropController/Search?type=Row
+        public async Task<ActionResult> Search([FromQuery] string type, [FromQuery] string query = "")
```
jSylvestre, line 147: Can't have it here. This entire class if FM access.

jSylvestre (changes requested):

Harvest.Web/Controllers/CropController.cs
srkirkland: yea you might be able to override w/ a new `[Authorize]` attribute which changes the policy to just any authorized user and still keep it in the class, but yup it needs to be changed.

Harvest.Web/ClientApp/src/Requests/Crops.tsx
sprucely: No, but I didn't like that options were being stored as type `string | { label: string }`. Which reminds me, I forgot to remove that union from the onSelect parameter.

Harvest.Web/Controllers/CropController.cs
sprucely: Removed FM from class and applied to all methods but this one.

Harvest.Web/ClientApp/src/Requests/Crops.tsx
sprucely: Nevermind. Just realized that the funky union type it was showing was only due to the data being passed in. Making the generic type explicit is redundant, so I removed it.

PR #565, opened 2021-09-22
Make sure projectId gets passed into DoCloseout

sprucely:
I verified that projectId now gets passed in.


jSylvestre (approved):

srkirkland (approved):
Looks good how did it possibly work before? Guess it doesn't matter.

sprucely:
Probably didn't actually test clicking the button, since I was primarily testing via the invoice job.

PR #567, opened 2021-09-23
Invoice path

### LINKED ISSUE #566: new invoice URL needs to be updated in sloth and possibly emails
URL is no longer like https://harvest-test.azurewebsites.net/Invoice/Details/9 since it needed to include the projectId, so now it's more like https://harvest-test.azurewebsites.net/invoice/details/3/9.

jSylvestre:
Resolve #566

### DIFF
--- a/Harvest.Core/Services/SlothService.cs
+++ b/Harvest.Core/Services/SlothService.cs
@@ -71,7 +71,7 @@ namespace Harvest.Core.Services
             var model = new TransactionViewModel
             {
                 MerchantTrackingNumber = invoiceId.ToString(),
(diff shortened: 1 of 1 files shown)


jSylvestre:
https://sloth-test.ucdavis.edu/harvest/Transactions/Details/a3f4424d-9b66-4ed3-98a1-521ffef78d24

srkirkland (approved):

srkirkland:
@jSylvestre do we not have an email which goes out for invoices and has a link to the invoice?  not saying we should, I just don't remember what does emails.

jSylvestre:
@srkirkland Currently we do not email when an invoice is generated.

PR #568, opened 2021-09-23
Added Date picker for filtering

### LINKED ISSUE #530: Fancy date filtering in tables
On the project table, I'd like it so filtering the start and end dates uses a datepicker.  Clicking on the a filter should bring up a datepicker where you can select a range of dates, and then the table will filter so that only that range is visible.  Will need to be easy to clear back to view all as well.

wongband:
resolve #530

### DIFF
--- a/Harvest.Web/package-lock.json
+++ b/Harvest.Web/package-lock.json
@@ -1,5 +1,13 @@
 {
   "name": "harvest.web",
   "version": "1.0.0",
-  "lockfileVersion": 1
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "harvest.web",
+      "version": "1.0.0",
+      "license": "MIT"
+    }
+  }
 }
--- a/Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/ProjectTable.tsx
@@ -6,9 +6,12 @@ import { Progress } from "reactstrap";
 import { ReactTable } from "../Shared/ReactTable";
 import { ReactTableUtil } from "../Shared/TableUtil";
 import {
+  DatePickerFilter,
   SelectColumnFilter,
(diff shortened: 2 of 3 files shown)


jSylvestre (changes requested):
Clearing the end date filter appears to hide the entire table contents.
Also, the end date filter should include dates that end on the date selected.

Harvest.Web/package-lock.json
```diff
@@ -1,5 +1,13 @@
 {
   "name": "harvest.web",
   "version": "1.0.0",
-  "lockfileVersion": 1
+  "lockfileVersion": 2,
```
srkirkland, line 4: Not sure this really makes sense, Harvest.Web doesn't have packages and shouldn't need a packages section in the package-lock.json file.

srkirkland (changes requested):
I think you have a good start and the code that's there looks good.  But I don't think it makes a sense what is being filtered right now, what I was looking for in the original issue was that each filter would allow you to select a range of dates.  So for example, clicking on the start filter would bring up something where you could say show me everything that starts between October 1st and Nov 1st and then it would filter to that.

Harvest.Web/package-lock.json
wongband: hmmm strange, I don't remember adding that but I will remove it

wongband:
@srkirkland, is there a specific way on how you want the filter to look like or could I just put two datepickers right under each other?

srkirkland:
> @srkirkland, is there a specific way on how you want the filter to look like or could I just put two datepickers right under each other?

@wongband our datepicker has a built in method for specifying ranges.  see https://reactdatepicker.com/#example-date-range-for-one-datepicker

srkirkland:
Or for your case, you probably want https://reactdatepicker.com/#example-date-range-using-input-with-clear-button

wongband:
@srkirkland ooo awesome, I will take a look at it.

PR #580, opened 2021-09-24
Fixed quote rejection

### LINKED ISSUE #532: Reject Quote Reason
Value isn't being passed

wongband:
resolve #532


srkirkland (commented):
I don't know of a better way but I'm going to look at this a bit more before merging, I don't like the idea of a POST inside a useEffect and a state variable as the trigger for the whole thing.  Doesn't seem properly reactive.  Hold off on merging for now.

PR #583, opened 2021-09-24
Jcs/rate passthrough use

jSylvestre:


jSylvestre:
Too much junk. Started over in other pr

PR #586, opened 2021-09-27
Allow approving quote with no acreage fee

### LINKED ISSUE #576: Ability to approve/route project with no acreage fee
APPROVE breaks because of an FK constraint error if the project have NO acreage fee.

Additionally, quotes look weird with 0 acreage fee.  handle it better.

sprucely:
Resolves #576

### DIFF
--- a/Harvest.Core/Models/QuoteModel.cs
+++ b/Harvest.Core/Models/QuoteModel.cs
@@ -34,7 +34,7 @@ namespace Harvest.Core.Models
         public string ProjectName { get; set; }
         public double Acres { get; set; }
         public double AcreageRate { get; set; }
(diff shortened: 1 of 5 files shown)


Harvest.Core/Models/QuoteModel.cs
```diff
@@ -79,5 +79,6 @@ public partial class WorkItem
         public double Total { get; set; }
         public string Unit { get; set; }
         public bool Markup { get; set; }
+        public bool IsPassthrough { get; set; }
```
jSylvestre, line 82: I thought I already added this...

jSylvestre (approved):
I didn't run an invoice through, but I think that should be ok. IIRC, it checks if acres is zero.

PR #587, opened 2021-09-27
Don't create $0.00 acreage expense

### LINKED ISSUE #549: Closeout acreage bug
I'm going to add a server side check, but FYI, this saves a zero amount in an expense:

image

sprucely:
Resolves #549


jSylvestre (changes requested):
It wasn't in the issue, but we should hide or disable the acreage part of the closeout if there are zero acres.

jSylvestre (changes requested):
If you try to save a quote with a acreage rate, but no acres, you get the error msg and save is disabled.
If you then clear out the acreage rate, the save is not enabled.
You either have to refresh, or select a rate, enter 1 acre and then unselect the rate.

jSylvestre (approved):
Has issues, but good enough for now.

PR #598, opened 2021-09-29
replace <a> with react router's <Link> element

### LINKED ISSUE #564: Fix link from invoice table
Right now it's using an `<a>` and doing a full refresh. We need it to use a react router <Link>

mxthu313:
resolve #564 
replace `<a>` with react router's `<Link>` element

### DIFF
--- a/Harvest.Web/ClientApp/src/Invoices/InvoiceTable.tsx
+++ b/Harvest.Web/ClientApp/src/Invoices/InvoiceTable.tsx
@@ -1,4 +1,5 @@
 import { useMemo } from "react";
+import { Link } from "react-router-dom";
 import { Cell, Column, TableState } from "react-table";
 import { ReactTable } from "../Shared/ReactTable";
 import { ReactTableUtil } from "../Shared/TableUtil";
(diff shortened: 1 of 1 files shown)


Harvest.Web/ClientApp/src/Invoices/InvoiceTable.tsx
```diff
@@ -16,9 +17,9 @@ export const InvoiceTable = (props: Props) => {
       {
         Cell: (data: Cell<Invoice>) => (
           <div>
-            <a href={`/invoice/details/${data.row.original.projectId}/${data.row.original.id}`}>
+            <Link to={`/invoice/details/${data.row.original.id}`}>
```
jSylvestre, line 20: @mxthu313 You need the projectId in here too

mxthu313: thanks for the catch.  sorry about that, fingers must have slipped.

jSylvestre (approved):
Looks good.
Unless there is something else you are changing in the PR, you can switch it Ready For Review and merge it.

PR #600, opened 2021-09-29
Wrap it in a try catch.

jSylvestre:


sprucely (approved):
This is fine, but I don't think execution is even getting to this.

PR #605, opened 2021-09-30
Used different react datepicker for more utility

### LINKED ISSUE #530: Fancy date filtering in tables

wongband:
resolve #530

Might need to change the styling of the date picker.


sprucely (approved):
Looks good. Might want to add a popup or something explaining how to select date ranges on the project date filters. It wasn't intuitive at first how to do it.

jSylvestre (changes requested):
Why is the package-lock so large?

wongband:
@jSylvestre I'm not too sure why. How come we don't gitignore our package-lock.json files btw?

sprucely:
@wongband What version of npm are you using? I'm on 6.14.15

> 
> @jSylvestre I'm not too sure why. How come we don't gitignore our package-lock.json files btw?

If you don't commit it, then the version of the application everyone else will get is different than what you are running locally. This means that things might work for you, but break on the CI/production/other local machines.

wongband:
@sprucely I have npm 7.20.5

jSylvestre (approved):

PR #606, opened 2021-09-30
Revalidate fields with errors on any change

### LINKED ISSUE #592: Revalidate fields that have failed validations on any change
Since field validations can now take into account the state of other fields, there needs to be a way to revalidate fields with errors when another field changes. The simplest approach would be to do a validateAll in an input's change event if there are currently any errors in the form.

sprucely:
resolves #592


jSylvestre (approved):
Didn't test it, but looks good to me.

PR #609, opened 2021-10-04
Fix broken tests

jSylvestre:
Resolve #607


jSylvestre:
@sprucely The developer who wrote the original test helper was sloppy with his dates :)

sprucely (approved):

PR #610, opened 2021-10-05
Jcs/rate object code

jSylvestre:
I don't know that we want to go this way.
Another way we could do it is change the billing unit to a dropdown then base the object code off of RR, PLS, or CAES. And PassThrough will have separate values

If we do this, we should reset the DB after deployment. Also, manually set the see value to invoices to 1000 in test to avoid future sloth duplicates

### DIFF
--- a/Harvest.Core/Domain/Expense.cs
+++ b/Harvest.Core/Domain/Expense.cs
@@ -59,6 +59,7 @@ namespace Harvest.Core.Domain
         [StringLength(50)]
         public string Account { get; set; }
 
+        //This is actually required, but that will make the migration a pain...
         public string ExpenseObjectCode { get; set; }
 
         internal static void OnModelCreating(ModelBuilder modelBuilder)
--- a/Harvest.Core/Domain/Expense.cs
+++ b/Harvest.Core/Domain/Expense.cs
@@ -60,6 +60,7 @@ namespace Harvest.Core.Domain
         public string Account { get; set; }
 
         //This is actually required, but that will make the migration a pain...
+        [StringLength(5)]
         public string ExpenseObjectCode { get; set; }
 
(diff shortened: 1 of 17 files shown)


Harvest.Core/Domain/Expense.cs
```diff
@@ -59,6 +59,9 @@ public class Expense
         [StringLength(50)]
         public string Account { get; set; }
 
+        //This is actually required, but that will make the migration a pain...
+        public string ExpenseObjectCode { get; set; }
```
sprucely: After generating the migration, it is okay to manually edit the corresponding code to first create the column as nullable, update with a default value, then change it to not nullable. But yeah, it is annoying.

sprucely (approved):

Harvest.Core/Domain/Expense.cs
sprucely: oops. Just noticed it's just a draft.

jSylvestre: The update would be a pain. This will need to wipe the DB for proper values first.
And I'll just make sure this is set with code.

srkirkland:
@jSylvestre I haven't actually read through this PR yet but from first sight we shouldn't be adding anything specifically for ObjectCode.  That ties us too much to KFS which is going away soon.  I don't know if AE will even have them.  I think it's better to fit all of the account info into the account field.  I have some docs I need to read through about account formats that include obj.

jSylvestre:
Which is why this is better :)
It isn't tied to the account, it is a separate field we can ignore in the future, also, it isn't for this account, but the account it comes from.

jSylvestre:
This will not be used, but keeping the code around as an example

PR #611, opened 2021-10-05
Maybe sloth transactions?

jSylvestre:
No idea if this is correct.


sprucely:
Looks okay to me. With the rounding stuff, did we address the possibility of being off by $0.01? I feel like we did, but don't recall any details.

jSylvestre:
> Looks okay to me. With the rounding stuff, did we address the possibility of being off by $0.01? I feel like we did, but don't recall any details.

Should catch that. I think I'm going to try and clean this up a little though so it is easier to follow.

jSylvestre:
replaced with other pr
Not Merged

PR #612, opened 2021-10-06
Jcs/sloth transaction grouping v2

jSylvestre:


Harvest.Core/Services/SlothService.cs
```diff
@@ -81,115 +81,29 @@ public async Task<Result<SlothResponseModel>> MoveMoney(int invoiceId)
                 return Result.Error("No expenses found for invoice: {invoiceId}", invoiceId);
             }
 
-            if (!invoice.Expenses.All(e => e.Total > 0))
+            var grandTotal = Math.Round(invoice.Expenses.Select(a => a.Total).Sum(),2);
+            if (grandTotal <= 0)
             {
+                Result.ResultError xx1x = Result.Error("Expenses found with a Total of 0 or less for invoice: {invoiceId}", invoiceId);
                 return Result.Error("Expenses found with a Total of 0 or less for invoice: {invoiceId}", invoiceId);
```
sprucely, line 88: This is weird. Why the extra call and variable?

Harvest.Core/Services/SlothService.cs
```diff
@@ -259,6 +173,127 @@ public async Task<Result<SlothResponseModel>> MoveMoney(int invoiceId)
...
+            var validatedProjectAccounts = new Dictionary<Account, AccountValidationModel>();
+            //Validate accounts
+            foreach (var projectAccount in invoice.Project.Accounts)
+            {
+                var account = await _financialService.IsValid(projectAccount.Number);
+                if (!account.IsValid)
+                {
+                    Result.Error("Unable to validate debit account {debitAccount}: {debitMessage}",
+                        projectAccount.Number, account.Message);
+                    return false;
```
sprucely, line 193: Not a big deal, but why not have this method return a `Task<Result<bool>>` and return these errors? Then the calling method could just pass along the error instead of creating its own `"ProcessDebits failed"` message.

Harvest.Core/Services/SlothService.cs
jSylvestre: Oops. Was trying to figure out a good way to do errors. I'll remove

Harvest.Core/Services/SlothService.cs
jSylvestre: I wasn't sure what that would look like to get it to log the error and return out of the calling method.

PR #615, opened 2021-10-07
PI Search Box Interaction

### LINKED ISSUE #573: PI Search box
IF PI selected, disable input. Allow clicking clear button which should re-enable.

Hint text and custom search text saying search by email or kerb.

mxthu313:
closes #573 

got around not being able to see the close button when disabled by using an AUX

image

8.10.2021 update:
image
image


mxthu313:
> Looks good to me. My only suggestion would be that you might want to create a class for the styles instead of using it directly. (In the SASS folder I think)

Will do 👍 . I actually thought about that too, but I decided to ultimately add the styling inline because it is a singular instance where this would be necessary.

srkirkland (approved):

PR #616, opened 2021-10-07
Basic save and validate of object code

### LINKED ISSUE #614: Changes to accounting, object codes, etc
I think fairly minor changes, mostly what we expect.  Since chart is going away and accounts are going to be more complicated with aggie enterprise in a year, we shouldn't add specific fields for parts of account identifiers.  We should just make sure the account number/text field is long enough, at least 50 chars.  Each rate will have an account string which includes OBJ code (and whatever it will be called in the future).  We will need a parser which can read a string and handle the components, and whatever front-end is easiest for entry.  something like `C-ACCT-SUBACCT-OBJ` format or similar.

On the PI side, they just enter their account like normal and depending on if the rate is passthrough or not we insert the correct OBJ code.

jSylvestre:
Resolve #614

### DIFF
--- a/Harvest.Web/Views/Rate/_RateForm.cshtml
+++ b/Harvest.Web/Views/Rate/_RateForm.cshtml
@@ -27,7 +27,7 @@
 </div>
 <div class="form-group">
     <label asp-for="Rate.Account" class="control-label"></label><span title="Account format: chart-account-{subaccount}-expenseObjectCode"> <i class="fas fa-question-circle fa-xs" ></i></span>
-    <input asp-for="Rate.Account" class="form-control" />
+    <input asp-for="Rate.Account" class="form-control" placeholder="x-xxxxxxx-xxxxx-xxxx"/>
     <span asp-validation-for="Rate.Account" class="text-danger"></span>
 </div>
 <div class="form-group">


Harvest.Core/Models/FinancialAccountModels/KfsAccount.cs
```diff
@@ -22,24 +22,25 @@ public class KfsAccount
...
             {
-                if (!string.IsNullOrWhiteSpace(Project))
-                {
-                    extraAccountInfo = $"-{SubAccount}-{Project}";
-                }
-                else
-                {
-                    extraAccountInfo = $"-{SubAccount}";
-                }
+                return $"{ChartOfAccountsCode}-{AccountNumber}-{SubAccount}-{ObjectCode}";
```
sprucely, line 34: What if `SubAccount` is null or whitespace here?

sprucely (dismissed):

Harvest.Core/Data/DbInitializer.cs
```diff
@@ -145,7 +145,7 @@ private async Task CreateEquipmentRates(User createdBy)
             rate.Type = Rate.Types.Equipment;
             rate.Description = "PLS Row Planter";
             rate.BillingUnit = "Plant Sciences";
-            rate.Account = "3-APSNFDS";
+            rate.Account = "3-APSNFDS--RAPB";
```
srkirkland, line 148: It's worth asking Bryan about the preferred format here because I've seen it written where if you don't have a subacct you have to put in 5 dashes (or however long the subacct is).  If that's the case our parser would have to handle that.

srkirkland (approved):
Looks good to me.  And even though I mention the format I think it's worth merging and we can update the parsing code if we find out more info about it later.

PR #618, opened 2021-10-11
Switch from monthly to yearly acreage Expenses

### LINKED ISSUE #613: Acreage fees upfront instead of prorated
For multi-year projects, we should charge yearly amount each year (on anniversary of project start, or close to it).

If a change request happens, we need to credit or pro-rate appropriately on the next invoice.  ex: if we go from 2 to 5 acres, we need to immediately create an expense for 3 new acres * acreagefee.

jSylvestre:
resolve #613

### DIFF
--- a/Harvest.Core/Services/ExpenseService.cs
+++ b/Harvest.Core/Services/ExpenseService.cs
@@ -49,7 +49,7 @@ namespace Harvest.Core.Services
             //So, if I look back 1 year, less a week (or maybe 14 days), that will not find the one at the start of the month.
             //If it was created Jan 3, and this is running Dec 1-7, it will find it and not create the fee
             var compareDate = now.Date.AddYears(-1).AddDays(7); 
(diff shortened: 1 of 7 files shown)


Harvest.Core/Services/ExpenseService.cs
```diff
@@ -16,87 +16,96 @@ namespace Harvest.Core.Services
...
-                throw new Exception("Amount must be >= 0.01");
+                return;
             }
-            var project = await _dbContext.Projects
-                .Include(p => p.AcreageRate)
-                .SingleAsync(p => p.Id == projectId);
 
-            var expense = await CreateExpense(project, amount);
+            //Look for an acreage expense in the last year or so.
+            var compareDate = DateTime.UtcNow.Date.AddYears(-1).AddDays(-1);
```
sprucely, line 43: Should AddDays() vary from -1 to -3, depending on weekend/business day?

PR #631, opened 2021-10-19
Jcs/refunds20211018

jSylvestre:


Harvest.Core/Services/SlothService.cs
```diff
@@ -94,6 +94,14 @@ public async Task<Result<SlothResponseModel>> MoveMoney(int invoiceId)
                 invoice.Total = grandTotal;
             }
 
+            if (invoice.Expenses.Any(a => a.Total < 0))
+            {
+                if (!(ProcessRefunds(model, grandTotal, invoice)).Value)
+                {
+                    return Result.Error("ProcessRefunds Failed");
+                }
```
srkirkland, line 102: I think this will swallow the error -- perhaps it's better to get back the Result<bool> and if it's .Value is false instead of creating a new error just return the error that you get back?

Harvest.Core/Services/SlothService.cs
```diff
@@ -165,10 +173,92 @@ public async Task<Result<SlothResponseModel>> MoveMoney(int invoiceId)
             return Result.Error("Sloth Response didn't have a success code for moneyTransfer id {transferId}: {data}", invoice.Id, data);
         }
 
+        private Result<bool> ProcessRefunds(TransactionViewModel model, decimal grandTotal, Invoice invoice)
+        {
+            var refundAmount = invoice.Expenses.Where(a => a.Total < 0).Sum(a => a.Total);
+            Log.Information("Refund amount should be negative here {refundAmount}", refundAmount);
```
srkirkland, line 179: I think it's worth actually checking w/ an if statement.  If there is nothing to refund then we should just exit early.

Harvest.Core/Services/SlothService.cs
srkirkland: ah so result.Error as well as being a return value also immediately logs?  interesting didn't know that.

Harvest.Core/Services/SlothService.cs
jSylvestre: I'll fix in a sec.

srkirkland (approved):

PR #639, opened 2021-10-21
Move API calls to /api namespace, default react otherwise

### LINKED ISSUE #271: Move API methods to /api/ url namespace
Probably via controller attributes, like `[Route("[controller]/[action]")]`?  might be easier than worrying too much about the names.

srkirkland:
We only had one controller that wasn't in the API folder, so I moved that one and now all except a few MVC controllers are using the new "API" route.  Any route in our app that doesn't start with /api now automatically returns the react page, so we no longer need to add placeholder react actions.  This does mean that react is now handling our 404s, which I think is fine.  If we want to handle API 404s differently we still can, but page 404s are react-router.

I think I've updated all the API calls, custom routes via attributes, and other edge cases, but I'll need to test more to be sure.

One thing to note is the webpack sockets connection was failing in dev for the same reason it used to fail when we added a 404 -- the fallback was catching it.  So I do a semi-hacky thing and our react fallback only fires on pages that don't start with `sockjs-node`.  I don't know if there's a better way, and it's not that bad.

closes #271

### DIFF
--- a/Harvest.Web/Views/Shared/_Layout.cshtml
+++ b/Harvest.Web/Views/Shared/_Layout.cshtml
@@ -88,7 +88,7 @@
                     @if (await UserService.HasAccess(AccessCodes.WorkerAccess))
                     {
                         <li class="nav-item">
-                            <a asp-controller="Expense" asp-action="Entry" class="nav-link">Expenses</a>
+                            <a href="/expense/entry" class="nav-link">Expenses</a>
                         </li>
                     }
                     <li class="nav-item">
--- a/Harvest.Web/ClientApp/src/App.tsx
+++ b/Harvest.Web/ClientApp/src/App.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { Route } from "react-router-dom";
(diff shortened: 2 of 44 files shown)


Harvest.Web/Views/Shared/_Layout.cshtml
```diff
@@ -88,7 +88,7 @@
                     @if (await UserService.HasAccess(AccessCodes.WorkerAccess))
                     {
                         <li class="nav-item">
-                            <a asp-controller="Expense" asp-action="Entry" class="nav-link">Expenses</a>
+                            <a href="/expense/entry" class="nav-link">Expenses</a>
```
sprucely, line 91: I think this will still cause a refresh. Maybe call history.push in `onClick` handler and cancel default behavior?

srkirkland: Yea that's ok, the change wasn't to help with the refresh, it was because the controller route would no longer work. Since the page is served from outside of react I think the full refresh is fine and matches the behavior of all the other links up there.

PR #672, opened 2021-10-29
hide orig fields

### LINKED ISSUE #671: Change request quote shows field location as already taken
The original project has the same field and is being counted as using the spot, even though it shouldn't count for the change request.

we probably need to filter out where field.project.id != originalproject.id

jSylvestre:
resolve #671

### DIFF
--- a/Harvest.Web/Controllers/Api/FieldController.cs
+++ b/Harvest.Web/Controllers/Api/FieldController.cs
@@ -22,12 +22,12 @@ namespace Harvest.Web.Controllers.Api
         // Get all fields that are part of projects active between the given dates
         public async Task<IActionResult> Active(DateTime start, DateTime end, int projectId)
         {
-            var project = await _dbContext.Projects.SingleAsync(a => a.Id == projectId);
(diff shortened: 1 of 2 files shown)


Harvest.Web/Controllers/Api/FieldController.cs
```diff
@@ -20,15 +20,18 @@ public FieldController(AppDbContext dbContext)
         }
 
         // Get all fields that are part of projects active between the given dates
-        public async Task<IActionResult> Active(DateTime start, DateTime end)
+        public async Task<IActionResult> Active(DateTime start, DateTime end, int projectId)
         {
+            var project = await _dbContext.Projects.SingleAsync(a => a.Id == projectId);
```
sprucely, line 25: If you only need OriginalProjectId, it's a bit cheaper to do...
```csharp
var originalProjectId = await _dbContext.Projects
    .Where(a => a.Id == projectId)
    .Select(a => a.OriginalProjectId)
    .SingleAsync();
```

sprucely (approved):

PR #693, opened 2021-11-02
Add confirmation to project closeout flow

### LINKED ISSUE #685: PI needs to approve closeout
Closeout process should put project into new state pending PI approval.  Email goes out, then when PI approves the project is actually closed out via whatever process happened before.  Projects in closeout state probably shouldn't be able to have expenses entered and should only route to approval is the amount of money doesn't put the project over budget.

sprucely:
resolves #685

WIP. Still need to ensure expenses can't be added by other containers and in backend when project status is `PendingCloseoutApproval`. Oh, and forgot to check tests as usual.

### DIFF
--- a/Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
@@ -127,7 +127,7 @@ export const UnbilledExpensesContainer = (props: Props) => {
           <div className="col text-right">
             <ShowFor
               roles={["FieldManager", "Supervisor", "Worker"]}
-              condition={project?.status !== "PendingCloseoutApproval"}
+              condition={project.status !== "PendingCloseoutApproval"}
             >
               <Link
                 to={`/expense/entry/${projectId}?${ExpenseQueryParams.ReturnOnSubmit}=true`}
--- a/Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
@@ -20,6 +20,7 @@ interface RouteParams {
 interface Props {
   newExpenseCount?: number; // just used to force a refresh of data when new expenses are created outside of this component
   hideProjectHeader?: boolean;
+  disableEdits?: boolean;
 }
 
 export const UnbilledExpensesContainer = (props: Props) => {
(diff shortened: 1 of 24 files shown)


jSylvestre:
Looks good so far. I'll test again once the other stuff is integrated

srkirkland (commented):
Looks good but I think you need to add some more history tracking to the action of initiating closeout.

Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
```diff
@@ -124,7 +125,10 @@ export const UnbilledExpensesContainer = (props: Props) => {
             )}
           </div>
           <div className="col text-right">
-            <ShowFor roles={["FieldManager", "Supervisor", "Worker"]}>
+            <ShowFor
+              roles={["FieldManager", "Supervisor", "Worker"]}
+              condition={project?.status !== "PendingCloseoutApproval"}
```
srkirkland, line 130: Why `project?.status`?  We've already checked to see that project !== undefined or else we wouldn't be showing any of this html at all

Harvest.Web/Controllers/Api/ExpenseController.cs
```diff
@@ -33,16 +33,23 @@ public ExpenseController(AppDbContext dbContext, IUserService userService, IProj
...
-                expense.CreatedBy     = user;
-                expense.CreatedOn     = DateTime.UtcNow;
-                expense.ProjectId     = projectId;
-                expense.InvoiceId     = null;
-                expense.Account       = allRates.Single(a => a.Id == expense.RateId).Account;
+                expense.CreatedBy = user;
+                expense.CreatedOn = DateTime.UtcNow;
+                expense.ProjectId = projectId;
+                expense.InvoiceId = null;
+                expense.Account = allRates.Single(a => a.Id == expense.RateId).Account;
```
srkirkland, line 52: What do you have against tab alignment?

srkirkland (approved):

jSylvestre (approved):
Didn't test the call to sloth for the final invoice pending, but everything else looked good.

Harvest.Web/ClientApp/src/Expenses/UnbilledExpensesContainer.tsx
sprucely: Didn't notice the `return <div>Loading...</div>;` part.

Harvest.Web/Controllers/Api/ExpenseController.cs
sprucely: Nothing. Should I disable autoformat for c# code?

PR #701, opened 2021-11-02
add key to home list

srkirkland:


jSylvestre (approved):
Didn't test, but seems fine.

PR #704, opened 2021-11-03
Hide empty project actions

sprucely:
Turns out calling `ShowFor` and `ShowForPiOnly` as functions wasn't too bad. I gave them aliases `useFor` and `useForPiOnly` to help remind that rules of hooks apply when using them as functions.


srkirkland (approved):
This is screaming to be extracted into a separate ProjectDetailActions component.  Otherwise though I think it looks good

PR #705, opened 2021-11-04
Fix build and eslint warnings

sprucely:
The `= null!` thing is necessary to satisfy C# 8's new nullable-reference warning.

### DIFF
--- a/Harvest.Core/Services/EmailService.cs
+++ b/Harvest.Core/Services/EmailService.cs
@@ -212,9 +212,10 @@ namespace Harvest.Core.Services
             return await QuoteDecision(project, reason, false);
         }
         [Obsolete("This was not implemented. If it is, the email has to be updated with new style.")]
-        public async Task<bool> ApproveAccounts(Project project, string[] emails)
+        public Task<bool> ApproveAccounts(Project project, string[] emails)
         {
             throw new NotImplementedException();
+            /*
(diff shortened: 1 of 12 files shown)


Harvest.Core/Services/EmailService.cs
```diff
@@ -212,9 +212,10 @@ public async Task<bool> QuoteDenied(Project project, string reason)
             return await QuoteDecision(project, reason, false);
         }
         [Obsolete("This was not implemented. If it is, the email has to be updated with new style.")]
-        public async Task<bool> ApproveAccounts(Project project, string[] emails)
+        public Task<bool> ApproveAccounts(Project project, string[] emails)
         {
             throw new NotImplementedException();
+            /*
```
jSylvestre, line 218: If you have to comment out code, please use // 
It makes in more clear.

Harvest.Jobs.Core/JobBase.cs
```diff
@@ -8,7 +8,7 @@ namespace Harvest.Jobs.Core
 {
     public abstract class JobBase
     {
-        public static IConfigurationRoot Configuration { get; set; }
+        public static IConfigurationRoot Configuration { get; set; } = null!;
```
jSylvestre, line 11: I don't really like that just to hide a warning. But I'll leave it to @srkirkland for a 🧑‍⚖️  (judgement)

srkirkland (approved):

PR #706, opened 2021-11-04
Order before taking

### LINKED ISSUE #702: Take operator ordering
Make sure wherever we use Take() we order by something, probably just last updated or created dates
```
The query uses a row limiting operator ('Skip'/'Take') without an 'OrderBy' operator. This may lead to unpredictable results.
```

jSylvestre:
Resolve #702

### DIFF
--- a/Harvest.Web/Controllers/Api/ExpenseController.cs
+++ b/Harvest.Web/Controllers/Api/ExpenseController.cs
@@ -105,6 +105,7 @@ namespace Harvest.Web.Controllers.Api
             // get projects where user has entered an expense in the last month
             var projects = await _dbContext.Projects.AsNoTracking()
                 .Where(p => p.IsActive && p.Expenses.Any(e => e.CreatedById == user.Id && e.CreatedOn > DateTime.UtcNow.AddMonths(-1)))
+                .OrderByDescending(a => a.Expenses.Max(m => m.CreatedOn))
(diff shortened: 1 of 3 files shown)


Harvest.Web/Controllers/Api/ExpenseController.cs
```diff
@@ -105,6 +105,7 @@ public async Task<ActionResult> GetRecentExpensedProjects()
             // get projects where user has entered an expense in the last month
             var projects = await _dbContext.Projects.AsNoTracking()
                 .Where(p => p.IsActive && p.Expenses.Any(e => e.CreatedById == user.Id && e.CreatedOn > DateTime.UtcNow.AddMonths(-1)))
+                .OrderByDescending(a => a.Expenses.Max(m => m.CreatedOn))
```
jSylvestre, line 108: @srkirkland Does this look correct?

sprucely: That looks a little iffy. Probably best to inspect the generated sql. When debugging you can inspect the queryable's underlying query. Well, at least you can in Visual Studio. You can also call `.ToQueryString()` to get it.

srkirkland: yea, I wouldn't do this, even if it works.  The main point is just to order by "something" so the result is consistent.  I'd say best to just orderBy project created date.

PR #707, opened 2021-11-04
Only show accounts if they exist 

### LINKED ISSUE #678: Invoice accounts when status is created
A Closeout invoice is in the created status (or one that fails to process).
The accounts section needs to say something like pending as these get set when sloth processes the invoice.

jSylvestre:
(the will not for a invoice in the created status).
resolve #678

### DIFF
--- a/Harvest.Web/ClientApp/src/Invoices/InvoiceDisplay.tsx
+++ b/Harvest.Web/ClientApp/src/Invoices/InvoiceDisplay.tsx
@@ -61,7 +61,7 @@ export const InvoiceDisplay = (props: Props) => {
         );
       })}
 
-      {accounts !== undefined && accounts.length > 0 ? (
+      {accounts?.length > 0 && (
         <div className="card-wrapper gray-top mt-4">
           <div className="card-header">
             <h4 className="primary-color bold-font">Account(s)</h4>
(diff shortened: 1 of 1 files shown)


Harvest.Web/ClientApp/src/Invoices/InvoiceDisplay.tsx
```diff
@@ -61,25 +61,29 @@ export const InvoiceDisplay = (props: Props) => {
...
-              <Row key={"transfer_" + transfers.id}>
-                <Col xs="10" sm="10">
-                  <div>{transfers.account}</div>
-                </Col>
-                <Col xs="2" sm="2">
-                  ${formatCurrency(transfers.total)}
-                </Col>
-              </Row>
-            ))}
+      {accounts !== undefined && accounts.length > 0 ? (
```
srkirkland, line 64: Cleaner way to write this would be `{ accounts?.length > 0 && (...html...) }` or `{ (accounts && accounts.length > 0) && (...html... }`.  You only need the `bool ? </> : null` thing if you are actually going to need the else case, otherwise just use `&&` so it's easier to read.

srkirkland (changes requested):

srkirkland (approved):

PR #711, opened 2021-11-04
Jcs/ticket fixes20211104

jSylvestre:


sprucely (approved):

Harvest.Web/ClientApp/src/Util/DateHelper.ts
```diff
@@ -0,0 +1,7 @@
+export const EnsureUTC = (date: Date) => {
```
srkirkland, line 1: Nitpicky but do the function camel case, that's what we do in other places. 'ensureUTC()'

Harvest.Web/ClientApp/src/Util/DateHelper.ts
```diff
@@ -0,0 +1,7 @@
...
+  if (date.toString().includes("Z")) {
+    return new Date(date);
+  }
+
+  return new Date(date + "Z");
```
srkirkland, line 6: I don't have a much better suggestion but I don't love relying on the implicit string conversion aspect here. Perhaps do like date.toISOString() or one of those methods? If not we can figure it out later, but it would nice to have something that gives a solidly defined output that won't vary between browsers.

srkirkland (approved):

PR #714, opened 2021-11-05
Add admin menu items

jSylvestre:
image


srkirkland (approved):
Good but ideally these wouldn't use href because that causes a full page refresh. It was ok before because admin links were all to non react pages. Can still merge if you want but we should really be using react router links for links to any react pages.

jSylvestre:
@srkirkland I tried adding it as a <Link to ...> but, formatting, and if I open it from a MVC page, it adds it to the page:

image

PR #715, opened 2021-11-05
Don't initiate closeout when quote is exceeded

### LINKED ISSUE #679: Deal with closeout when expenses exceed quote
Right now, closeout fails to create the invoice (good ? ) but the message isn't clear why.
Maybe grey out closeout or have bigger better error message.

sprucely:
Resolves #679

Wasn't sure about getting total unbilled from child component. But I think it's better than making an extra api call.

### DIFF
--- a/Harvest.Core/Services/InvoiceService.cs
+++ b/Harvest.Core/Services/InvoiceService.cs
@@ -58,6 +58,13 @@ namespace Harvest.Core.Services
                 return Result.Error("Project status is not Active or AwaitingCloseout");
             }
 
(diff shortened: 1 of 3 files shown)


srkirkland (approved):
don't love the child component pass-up reporting either but I think it's a fine compromise at this point.  The problem is really that the `UnbilledExpensesContainer` is being used sometimes like a container and sometimes like a sub-component, when really it shouldn't and don't think we do this elsewhere.  A fix would require refactoring it to have a shared sub-component that was passed the unbilled expenses.  I'm not sure it's worth it, but I'll approve the PR and leave that up to you.

PR #723, opened 2021-11-15
Error page

jSylvestre:


sprucely (approved):
That's much better than an endless loop. I think it's fine to merge after tests are fixed.

jSylvestre:
(The endless loop did allow me to eventually log in )

srkirkland (approved):

PR #726, opened 2021-11-16
Get accurate decimal count for very small numbers

### LINKED ISSUE #725: Expense Bug
image

sprucely:
resolves #725

Tested all edge cases for `Calculations.countDecimals(num)` on TS playground, but currently can't login to test validation until CAS IAM issue is resolved.

### DIFF
--- a/Harvest.Web/ClientApp/src/Util/Calculations.ts
+++ b/Harvest.Web/ClientApp/src/Util/Calculations.ts
@@ -17,6 +17,23 @@ export const roundToTwo = (num: number) => {
   return Math.round(num * 1e2) * 1e-2;
 };
 
+export const countDecimals = (num: number) => {
+  if (Math.floor(num.valueOf()) === num.valueOf()) return 0;
+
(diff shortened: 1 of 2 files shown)


jSylvestre:
I'll let Scott decide, but I think we should make this more simple.
Make it invalid if it contains any non number or . character (E or e or -) and then just check the number of characters after the decimal

srkirkland:
@sprucely not following exactly what the issue was that prompted this, but I think `num.toFixed(2)` should do a lot of the heavy lifting for you here, maybe just check if num === num.toFixed(2)?

jSylvestre:
> @sprucely not following exactly what the issue was that prompted this, but I think `num.toFixed(2)` should do a lot of the heavy lifting for you here, maybe just check if num === num.toFixed(2)?

@sprucely @srkirkland need to compare like:
console.log(Number.parseFloat(1.1) === Number.parseFloat(Number.parseFloat(1.1).toFixed(2)));

PR #727, opened 2021-11-16
fix up responsiveness

jSylvestre:


sprucely (commented):
Is this how the responsive feature is supposed to behave? It doesn't start collapsing columns until content has already started scrolling off the page...
image

jSylvestre:
@sprucely Not sure, but that was the best I could get it. Other things I tried ended up not showing some info at all.

sprucely (approved):

PR #729, opened 2021-11-17
lookup IAM ID if CAS doesn't give it to us

srkirkland:


sprucely (approved):
Works for me.

Harvest.Web/Startup.cs
```diff
@@ -93,6 +93,23 @@ public void ConfigureServices(IServiceCollection services)
                         return;
                     }
 
+                    // Sometimes CAS doesn't return the required IAM ID
+                    // If this happens, we take the reliable Kerberos (NameIdentifier claim) and use it to lookup IAM ID
+                    if (!identity.HasClaim(c => c.Type == "ucdPersonIAMID"))
+                    {
+                        var identityService = context.HttpContext.RequestServices.GetRequiredService<IIdentityService>();
```
jSylvestre, line 100: Looks like CAS stage is fixed, but this is probably good to keep.
Throw in a log error message to let un know?

jSylvestre (changes requested):

jSylvestre (approved):

PR #730, opened 2021-11-17
Have a layout page with no react or auth stuff

jSylvestre:
This looks a little better (no react/menu)

image


srkirkland:
@jSylvestre you don't really need the layout2 page at all, just put it straight in that error page.  and then you can also leave off a bunch of those scripts as well if you want since you aren't using react or other dynamic stuff.

srkirkland:
if you do prefer a layout2 page rename it something like emptyLayout

jSylvestre:
@srkirkland  I had just copied the other layout and cut out stuff that didn't look like it was needed. I originally had it named the same, but I didn't want confusion.

But I can just dump it on the page like you had in your original suggestion.

jSylvestre:
image

sprucely (approved):
Looks good. But why the react scripts still included? Are they needed for the styles?

jSylvestre:
I didn't get the green styles without them.

Harvest.Web/Harvest.Web.csproj
```diff
@@ -45,6 +45,16 @@
...
+  <ItemGroup>
+    <None Remove="ClientApp\public\error.png" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="ClientApp\public\error.png">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
```
srkirkland, line 57: You don't need this -- see for example that this page references `/media/ucdavis_white_logo.svg` just fine and it's not in the csproj.  CRA will build and anything in the ClientApp/public folder will be made available at the root, nothing needed in csproj

srkirkland (changes requested):

srkirkland:
> I didn't get the green styles without them.

FWIW, only in development do you not get the green styles, in prod it'd work without the scripts.  But it doesn't really bug me too much that they are unnecessarily loaded on an error page.

PR #732, opened 2021-11-18
Cyd/sitecleanup 9

cydoval:
lots of responsive tweaks and tune ups to paddings/margins


jSylvestre (approved):
Most of the changes appear to be subtle.
Looks good to me.

PR #736, opened 2021-11-19
Possibly better way of getting dynamic scripts from CRA

srkirkland:
Previous approach involved reading the contents of the scripts/styles directories.  New way just reads the compiled index.html file and spits out the relevant sections.  I think it's probably a better approach because we'll be guaranteed to get the ordering right, and any future changes to build chunks should automatically update.

I do feel a little dirty about reading the contents of a file as a string and then doing regex over HTML, but it's a known file and it's pretty small, plus we cache the output so I don't see the harm.


sprucely (approved):
I like this. Seems less likely to break on CRA upgrades.

jSylvestre (approved):
Only think different I noticed was this warning in the console:
`webpackHotDevClient.js:138 src\Closeout\CloseoutContainer.tsx
  Line 3:18:  'UncontrolledTooltip' is defined but never used  @typescript-eslint/no-unused-vars`

PR #738, opened 2021-12-03
Add antiforgery token to all API requests

### LINKED ISSUE #555: API Antiforgery
The API controllers just inherit from the Controller which doesn't have the autoAntiforgery attribute.
Do we want to change this. Probably a bunch of things will break.

srkirkland:
Updated all API controllers to inherit from SuperController, which means they are now behind `[AutoValidateAntiforgeryToken]`.  I then injected the token into the Layout.cshtml page and read it into react as part of our `Harvest` var just like user info.  Next I made a simple wrapper around `fetch()` with `authenticatedFetch()` and you need to use that for any non-GET method, though it's cleaner to just try to remember to use it everywhere you would have done fetch.  Similar idea to peaks except I didn't attach the new fetch stand-in to `context` which I think keeps it cleaner and more flexible.

One thing to note - the new `authenticatedFetch()` assumes a JSON payload accepting a JSON response.  AFAIK we use this for every fetch, but if in the future we have a reason to do something else we'll need to tweak it to allow header overrides or to directly expose the token or by some other means.  Probably won't come up but just in case....

closes #555

### DIFF
--- a/Harvest.Web/ClientApp/src/Closeout/CloseoutConfirmationContainer.tsx
+++ b/Harvest.Web/ClientApp/src/Closeout/CloseoutConfirmationContainer.tsx
@@ -12,6 +12,7 @@ import { useConfirmationDialog } from "../Shared/ConfirmationDialog";
 import { ShowFor } from "../Shared/ShowFor";
 import { useIsMounted } from "../Shared/UseIsMounted";
 import AppContext from "../Shared/AppContext";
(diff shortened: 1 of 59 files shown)


sprucely (approved):
Looks good and works for me. I'd say, once tests are fixed, 🚢.

jSylvestre (approved):
Only did minimal testing, but worked with the help ticket send.
🚢

PR #739, opened 2021-12-09
update all projects to .net6 and update nuget packages

srkirkland:
only code change needed was to update ticket.cs for proper location of jsonIgnore.

Seems to be working perfectly, but will need to double check all complex parts like email templating and file upload and other stuff just to make sure.

Startup is quicker and hot reloading is :fire:!


jSylvestre (changes requested):
Open an exiting ticket fails with:

`Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'map')
`

jSylvestre:
Didn't get the ticket error when running from code, but did still get the attachment error

jSylvestre (approved):

sprucely (approved):

PR #743, opened 2022-01-18
move script to head

jSylvestre:


sprucely (approved):
Does this cause it to not render the page until after the alert is dismissed?

srkirkland (changes requested):
stick it right before the body closing tag

PR #745, opened 2022-01-19
Show/Hide edit delete

jSylvestre:


Harvest.Web/Controllers/RateController.cs
```diff
@@ -34,6 +34,7 @@ public RateController(AppDbContext dbContext, IUserService userService, IFinanci
         public async Task<ActionResult> Index()
         {
             var rates = await _dbContext.Rates.Where(a => a.IsActive).ToListAsync();
+            ViewBag.AllowEdit = await _userService.HasAnyRoles(new string[] { Role.Codes.System, Role.Codes.FieldManager });
```
sprucely, line 37: I know we've used it in a lot of apps, but ViewBag has always felt dirty to me. Is it just so we can use a domain model and avoid having to create a ViewModel?

jSylvestre: Pretty much. But I'd say a model to do this would be overkill.

sprucely (approved):

srkirkland (approved):

PR #749, opened 2022-01-20
show error notification when project doesn't load

jSylvestre:
I can do this in other places as well, such as the quote container.


srkirkland (approved):
seems good to me

PR #754, opened 2022-01-21
Getting emails to format correctly with outlook.

jSylvestre:
InvoiceCreated so far.

Need to also replace the test logo with the prod one once that is deployed. Then regenerate the cshtml


srkirkland (approved):
seems good but can't really check out the mjml changes but I'm sure they are fine

PR #757, opened 2022-02-16
fix promise warning

jSylvestre:


srkirkland (commented):
I can't say for sure but I'm thinking there would be a better way to solve this that doesn't require the cast on every call.  Probably by returning something specific from the `responseMap()` method.  I'd want to investigate that before going this way

srkirkland (approved):

PR #764, opened 2022-03-29
Jcs/auto closeout job

### LINKED ISSUE #761: 18 day auto closeout
If awaiting closeout for that long, just auto-close

### LINKED ISSUE #762: Send daily reminders for awaiting closeout
BCC farm managers?

jSylvestre:
Closes #761 
Closes #762 

Completely untested....

### DIFF
--- a/Harvest.Core/Services/EmailService.cs
+++ b/Harvest.Core/Services/EmailService.cs
@@ -44,8 +44,10 @@ namespace Harvest.Core.Services
         Task<bool> InvoiceDone(Invoice invoice, string status);
         Task<bool> InvoiceError(Invoice invoice);
 
-        Task<bool> CloseoutConfirmation(Project project); //Project is awaiting PI to close it
+        Task<bool> CloseoutConfirmation(Project project, bool ccFieldManagers = true); //Project is awaiting PI to close it
         Task<bool> ProjectClosed(Project project); //Project has been closed by PI
+
+        Task<int> ResendCloseoutNotifications();
     }
 
     public class EmailService : IEmailService
@@ -575,14 +577,14 @@ namespace Harvest.Core.Services
             return true;
         }
 
-        public async Task<bool> CloseoutConfirmation(Project project)
+        public async Task<bool> CloseoutConfirmation(Project project, bool ccFieldManagers = true)
         {
             try
             {
                 var projectUrl = $"{_emailSettings.BaseUrl}/Project/Details/";
                 var closeoutUrl = $"{_emailSettings.BaseUrl}/Project/CloseoutConfirmation/";
                 var emailTo = new string[] { project.PrincipalInvestigator.Email };
-                var ccEmails = await FieldManagersEmails();
+                var ccEmails = ccFieldManagers ? await FieldManagersEmails() : null;
 
                 var model = new CloseoutConfirmationModel()
                 {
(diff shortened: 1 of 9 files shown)


Harvest.Core/Services/EmailService.cs
```diff
@@ -608,6 +610,24 @@ public async Task<bool> CloseoutConfirmation(Project project)
             }
 
             return true;
+        }
+
+        public async Task<int> ResendCloseoutNotifications()
```
srkirkland, line 615: probably should just be `SendCloseoutNotifications()` because the service itself doesn't know why we are sending them.

Harvest.Core/Services/EmailService.cs
```diff
@@ -608,6 +610,24 @@ public async Task<bool> CloseoutConfirmation(Project project)
...
+
+        public async Task<int> ResendCloseoutNotifications()
+        {
+            var count = 0;
+            var projectsAwaitingCloseout = await _dbContext.Projects.Include(a => a.PrincipalInvestigator).Where(a => a.IsActive && a.Status == Project.Statuses.PendingCloseoutApproval).ToArrayAsync();
+            if (projectsAwaitingCloseout.Any())
+            {
+                foreach(var project in projectsAwaitingCloseout)
+                {
+                    if(await CloseoutConfirmation(project, false))
```
srkirkland, line 623: Don't love error checking by returning true/false but i think the bigger issue is if one send fails it'll fail all of them.  I think a wrap in try/catch within the for loop here would be a good idea

Harvest.Core/Services/InvoiceService.cs
```diff
@@ -232,5 +234,29 @@ public async Task<List<int>> GetCreatedInvoiceIds()
...
+        /// Automatically closeout projects that are 18 days or older since closeout was initiated.
+        /// </summary>
+        /// <param name="daysBeforeAutoCloseout">Negative Value</param>
+        /// <returns></returns>
+        public async Task<int> AutoCloseoutProjects(int daysBeforeAutoCloseout)
+        {
+            var dateCutoff = DateTime.UtcNow.AddDays(daysBeforeAutoCloseout);
+
+            var count = 0;
+            var potentialProjectIds = await _dbContext.Projects.Where(a => a.IsActive && a.Status == Project.Statuses.PendingCloseoutApproval).Select(a => a.Id).ToArrayAsync();
```
srkirkland, line 248: Maybe not worth the effort but i don't like doing this via the history.  I think a `LastUpdateOn` db field would make things a lot cleaner here, and then we'd just update that flag when a project changes status.

Harvest.Core/Services/InvoiceService.cs
```diff
@@ -232,5 +234,29 @@ public async Task<List<int>> GetCreatedInvoiceIds()
...
+            var projectIds = await _dbContext.ProjectHistory.Where(a => a.Action == "ProjectCloseoutInitiated" && potentialProjectIds.Contains(a.ProjectId) && a.ActionDate < dateCutoff).Select(a => a.ProjectId).ToArrayAsync();
+
+            foreach (var projectId in projectIds)
+            {
+                var result = await CreateInvoice(projectId, true);
```
srkirkland, line 253: let's add some logging since this is a fairly sensitive thing, auto closeout.  also maybe a try/catch in case one project has issues

Harvest.Core/Services/InvoiceService.cs
jSylvestre: @srkirkland I'm ok with that, but that would be a new DB field

Harvest.Core/Services/EmailService.cs
jSylvestre: @srkirkland  The CloseoutConfirmation() is wrapped in a try catch, but I don't know why all would fail if one fails.

Harvest.Core/Services/InvoiceService.cs
srkirkland: @jSylvestre yea i was just thinking something simple like add a specific log saying something like "Initiating auto closeout for project 123"

Harvest.Core/Services/EmailService.cs
srkirkland: ok that'll work

Harvest.Core/Services/InvoiceService.cs
srkirkland: yea.  i think it'd be better though

srkirkland (approved):

PR #771, opened 2022-04-04
Use last action date.

jSylvestre:
Don't deploy until existing project has been auto closed


srkirkland (approved):

srkirkland:
IF you wanted you could just run a command to update lastUpdated on existing projects to today and i think it shouldn't break anything and should work for the autocloseout

PR #772, opened 2022-04-05
Jcs/adhoc project

jSylvestre:
image

### DIFF
--- a/Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
@@ -208,7 +208,8 @@ export const AdhocProject = () => {
     const response = await request;
 
     if (response.ok) {
-      history.push("/"); //TODO: Open project that was created instead
+      const data = await response.json();
+      history.push(`/Project/Details/${data.id}`);
     }
   };
 
--- a/Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
+++ b/Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
@@ -8,6 +8,7 @@ import {
   WorkItemImpl,
   Project,
   CropType,
+  ProjectAccount,
 } from "../types";
 import { ActivityForm } from "../Quotes/ActivityForm";
 import { Button, FormGroup, Input, Label } from "reactstrap";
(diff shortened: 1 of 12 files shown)


Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
```diff
@@ -0,0 +1,240 @@
...
+    // get rates so we can load up all expense types and info
+    const cb = async () => {
+      const response = await authenticatedFetch("/api/Rate/Active");
+
+      if (response.ok) {
+        const rates: Rate[] = await response.json();
+
+        getIsMounted() && setRates(rates);
+
+        // create default activity
```
srkirkland, line 77: ?

Harvest.Web/Controllers/Api/ProjectController.cs
```diff
@@ -185,5 +186,115 @@ public async Task<IActionResult> RefreshTotal(int projectId)
...
+                // Accounts will be auto-approved by quote approver
+                account.ApprovedById = currentUser.Id;
+                account.ProjectId= newProject.Id;
+                account.ApprovedOn = DateTime.UtcNow;
+                percentage += account.Percentage;
+                if (account.Percentage < 0)
+                {
+                    return BadRequest("Negative Percentage Detected");
+                }
+                newProject.Accounts.Add(account); //Don't need to specify projectId if I add it to the project?
```
srkirkland, line 264: I believe yes, as long as you are adding to the array you don't need to set projectId, and thus you don't need to do this double-save thing, you can just do it all at once

jSylvestre: Yeah... I had a different error at first and it wasn't working and I couldn't catch the exception. I'll change back.

Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
jSylvestre: @srkirkland  It is in the expense container:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Web/ClientApp/src/Expenses/ExpenseEntryContainer.tsx#L95

srkirkland (approved):

PR #773, opened 2022-04-06
Populate a quote

jSylvestre:


Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
```diff
@@ -58,6 +60,8 @@ export const AdhocProject = () => {
   const [accounts, setAccounts] = useState<ProjectAccount[]>([]);
   const [disabled, setDisabled] = useState<boolean>(true);
 
+  const [quote, setQuote] = useState<QuoteContent>(new QuoteContentImpl());
```
srkirkland, line 63: This should not be a state variable -- you aren't ever calling setQuote() and react doesn't need to know about the content.

Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
```diff
@@ -142,10 +146,13 @@ export const AdhocProject = () => {
         )
     );
 
+    quote.activities = activities.filter((a) => a.workItems.length > 0);
```
srkirkland, line 149: You should never set a property on a state variable itself, that's what the setter is for

Harvest.Web/Controllers/Api/ProjectController.cs
```diff
@@ -273,10 +273,51 @@ public async Task<IActionResult> CreateAdhoc([FromBody] AdhocPostModel postModel
...
             await _dbContext.SaveChangesAsync();
 
+            var quote = new Quote();
+            quote.InitiatedById = currentUser.Id;
+            quote.Status = "New"; // TODO: definte status progression
+            quote.CreatedDate = DateTime.UtcNow;
+            quote.Project = newProject;
+            quote.ProjectId = newProject.Id;
+            quote.Total = postModel.Expenses.Select(a => a.Total).Sum(); //Calculate totals here? probably not done yet
+            postModel.Quote.EquipmentTotal = (double)postModel.Expenses.Where(a=> a.Type == "Equipment").Select(a => a.Total).Sum();
```
srkirkland, line 286: do we not have statics/enums for these?

srkirkland (changes requested):

Harvest.Web/Controllers/Api/ProjectController.cs
jSylvestre: Yeah, Rate.Types

Harvest.Web/ClientApp/src/Projects/AdhocProject.tsx
jSylvestre: @srkirkland so :
 const quote: QuoteContent = {
      projectName: project.name,
      acres: 0,
      acreageRate: 0,
      years: 0,
      acreageTotal: 0,
      activitiesTotal: 0,
      laborTotal: 0,
      equipmentTotal: 0,
      otherTotal: 0,
      grandTotal: 0,
      fields: [],
      activities: activities.filter((a) => a.workItems.length > 0),
      acreageRateId: null,
      acreageRateDescription: ""
    }

srkirkland: sure or just use `const quote = new QuoteContentImp();` and then set activities from there, assuming it sets the defaults you want

PR #776, opened 2022-04-28
set minimum log level for elastic apm

srkirkland:


jSylvestre (approved):
Do we have that setting in any jobs?

sprucely:
@jSylvestre I don't believe I've seen APM configured for any jobs.

PR #792, opened 2023-01-10
pass in config setting

jSylvestre:


sprucely (approved):
Didn't test, but the code looks fine to me.

PR #795, opened 2023-01-17
Jcs/aggie enterprise sloth20230117

jSylvestre:


srkirkland (commented):
Sloth and sloth v2 are going to be the same URLs or base URLs anyway. For GS at least I'm just going to have one Url stored in settings and then the code will tack on v1/ or v2/ on the httpClient as needed.

Issue #801, opened 2023-02-02
Project Accounts - KFS Convert 


jSylvestre:
If project account is KFS, try to use the KFS convert.

PR #803, opened 2023-02-02
Update nuget packages.

jSylvestre:
Fixes the skip in theory tests

### DIFF
--- a/Harvest.Core/Services/SlothService.cs
+++ b/Harvest.Core/Services/SlothService.cs
@@ -1,3 +1,4 @@
+using AggieEnterpriseApi.Validation;
 using Harvest.Core.Data;
 using Harvest.Core.Domain;
 using Harvest.Core.Extensions;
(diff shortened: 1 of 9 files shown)


jSylvestre:
closes #801

Harvest.Core/Services/SlothService.cs
```diff
@@ -81,11 +82,24 @@ public async Task<Result<SlothResponseModel>> MoveMoney(int invoiceId)
...
             }
+
             if (_aeSettings.UseCoA)
             {
                 //Do an early check to make sure project account(s) are valid
                 foreach (var account in invoice.Project.Accounts)
                 {
+                    if(FinancialChartValidation.GetFinancialChartStringType(account.Number) == FinancialChartStringType.Invalid)
+                    {
+                        //Ok, it is an invalid COA format, so 99.99% chance this is still a KFS account. Lets try and convert it:
```
srkirkland, line 93: probably an ok assumption, but do you think it's worth checking that the format is `3-` or i guess `{alpha}-` first before assuming KFS?

srkirkland (approved):

PR #805, opened 2023-02-22
Give GL COAs our best guess at a description

jSylvestre:
Dept name and Fund Name

### DIFF
--- a/Harvest.Core/Models/FinancialAccountModels/AccountValidationModel.cs
+++ b/Harvest.Core/Models/FinancialAccountModels/AccountValidationModel.cs
@@ -21,7 +21,7 @@ namespace Harvest.Core.Models.FinancialAccountModels
 
         public string AccountManager { get; set; }
         public string AccountManagerEmail { get; set; }
-        public string Description { get; set; } //Description of the COA.
+        public string Description { get; set; } //Description of the COA. 
 
         /// <summary>
         /// Return Segment info.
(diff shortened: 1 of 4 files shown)


Harvest.Core/Services/AggieEnterpriseService.cs
```diff
@@ -95,7 +95,11 @@ public async Task<AccountValidationModel> IsAccountValid(string financialSegment
                 }
 
                 rtValue.GlSegments = FinancialChartValidation.GetGlSegments(financialSegmentString);
-
+                
+                rtValue.CoaDescription = $"{data.GlValidateChartstring.SegmentNames.DepartmentName} - {data.GlValidateChartstring.SegmentNames.FundName}";
```
sprucely, line 99: Are some department names kind of wordy? Should this be wrapped in some form of safe-truncate?

Harvest.Core/Models/FinancialAccountModels/AccountValidationModel.cs
```diff
@@ -21,7 +21,7 @@ public class AccountValidationModel
 
         public string AccountManager { get; set; }
         public string AccountManagerEmail { get; set; }
-        public string ProjectName { get; set; }
+        public string CoaDescription { get; set; }
```
srkirkland, line 24: Rather just name it description

Harvest.Core/Services/AggieEnterpriseService.cs
jSylvestre: Seems fine
image

image

sprucely (approved):

srkirkland (approved):

PR #806, opened 2023-02-23
Fix wording in log

jSylvestre:


sprucely (approved):
Well that certainly changes the meaning

PR #821, opened 2023-04-03
Stuff that needs to change for sloth teams

jSylvestre:
This will not compile yet, but should be type of changes needed when the team is attached to the project and has the sloth api key set.


jSylvestre:
superseded by #834

PR #823, opened 2023-04-05
Team routes for project & expenses

srkirkland:
Adds team routes to everything that needs it, like a list of all projects, projects needing attention, ticket lists, expenses, etc.  Rate + Permissions are non-MVC pages which are also now behind /team/xyz paths, though not yet implemented.

Nav is still wonky, it won't work if you are navigating from non-team pages back to team pages -- that'll be fixed later.


jSylvestre:
You loose the team if you click on a action that doesn't have it.
Then on something that should.

jSylvestre (approved):

PR #824, opened 2023-04-05
Start of implementing teams on the permissions controller

jSylvestre:


Harvest.Core/Data/DbInitializer.cs
```diff
@@ -82,15 +82,26 @@ public async Task Initialize(bool recreateDb)
...
             await _dbContext.SaveChangesAsync();
 
             return;
-        }
-
+        }
+
+        private async Task UpdateTeamPermissions()
+        {
+            var permissions = await _dbContext.Permissions.Include(a => a.Role).Where(a => a.TeamId == null && a.Role.Name != Role.Codes.System).ToListAsync();
```
srkirkland, line 98: why include role?

Harvest.Web/Controllers/PermissionsController.cs
```diff
@@ -15,13 +15,16 @@
...
 {
+
     [Authorize(Policy = AccessCodes.FieldManagerAccess)]
     public class PermissionsController : SuperController
     {
         private readonly AppDbContext _dbContext;
         private readonly IIdentityService _identityService;
         private readonly IUserService _userService;
 
+        public string TeamSlug => ControllerContext.RouteData.Values["team"] as string;
```
srkirkland, line 26: move to super controller since this will be used in multiple controllers eventually

srkirkland (approved):

PR #829, opened 2023-04-06
Fix internal routing consistency for client nav

srkirkland:
Now you can click around and all buttons should work as expected within project/quote/invoice/expense stuff.  Does not cover jobs (email).

All the tests will fail, //TODO


srkirkland:
Wow, tests actually passed...

PR #830, opened 2023-04-10
Team fields for Sloth

jSylvestre:


srkirkland (changes requested):
Since we are going `.Include(p => p.Team)` on most queries, I don't really want these things living along with the `Team`.  I haven't yet thought through the best way, but first thing that comes to mind is make a `TeamDetails` class/table and put them in there with a one-to-one mapping https://learn.microsoft.com/en-us/ef/core/modeling/relationships/one-to-one.

jSylvestre:
Don't merge

PR #837, opened 2023-04-12
Filter rates for team

jSylvestre:


srkirkland (approved):

Harvest.Web/Controllers/RateController.cs
```diff
@@ -55,10 +56,16 @@ public async Task<ActionResult> Index()
 
         // Returns the Active rates via API
         // GET: RateController/Active
+        [Route("api/{team}/{controller}/{action}")]
         public async Task<ActionResult> Active()
         {
-            //TODO: Use team to filter rates 
-            var rates = await _dbContext.Rates.Where(a => a.IsActive).OrderBy(a => a.Description).Select(r => new { r.Price, r.Unit, r.Type, r.Description, r.Id, r.IsPassthrough }).ToArrayAsync();
+            var team = await _dbContext.Teams.SingleOrDefaultAsync(t => t.Slug == TeamSlug);
```
srkirkland, line 62: `if (await _dbContext.Teams.Any(t => t.Slug == TeamSlug) == false) { return BadRequest() }` is a more efficient way of querying since it doesn't need to return the team.  i know the team is small but still.  Then when querying rate it's fine to do `.Where(a=> a.Team.Slug == TeamSlug ..... ` in there since it's indexed.  Or alternately just do the query for the teamId using `.Select(t => t.Id)` and you'll get the nullable int and can use it the same way you used the whole team

PR #838, opened 2023-04-12
Don't get the team unless really needed.

jSylvestre:
Also, show team on edit and delete pages


srkirkland (approved):
in this case I would just grab the whole team since you are using slug & name, but still I think this works and it's handy to have the name displayed so up to you if you want to do it this way or the old way.

PR #841, opened 2023-04-13
Don't really like this

jSylvestre:


sprucely (commented):
So how likely is it that a team name happens to match a controller action name?

srkirkland (commented):
What was wrong with using the built in api for getting route values?

jSylvestre:
> So how likely is it that a team name happens to match a controller action name?

Never. We will control teams and there will only be a few.

srkirkland (approved):

PR #845, opened 2023-04-14
Catch Project details if not authorized

jSylvestre:


Harvest.Web/ClientApp/src/Projects/ProjectDetailContainer.tsx
```diff
@@ -42,13 +42,15 @@ export const ProjectDetailContainer = () => {
...
       );
-
       if (response.ok) {
         const project = (await response.json()) as Project;
         getIsMounted() && setProject(project);
         setIsLoading(false);
       } else {
         setNotification(response, "Loading", "Error Loading Project");
+        if (response.status === 403) {
+          window.location.replace("/Account/AccessDenied");
```
srkirkland, line 52: I think we should be using `history.replace()` instead of `window` directly.

srkirkland (approved):

PR #846, opened 2023-04-14
Add a desc to the team

jSylvestre:


sprucely (commented):
Didn't test, but code looks fine.

srkirkland (approved):
I think I'd probably add to teamDetail instead of having it pull along with every .Include(a => a.Team). Not a huge deal so I'll leave it up to you if you want to change, but we shouldn't add more to team in general, other than maybe an active flag at some point.

jSylvestre:
@cydoval FYI, made the whole card clickable, hid the underline on hover and have a check from fontawesome.
But not sure how to change the icon...
image

PR #852, opened 2023-04-18
Jcs/finance role allow more things

jSylvestre:


Harvest.Core/Models/AccessCodes.cs
```diff
@@ -11,13 +11,18 @@ public static class AccessCodes
...
+        public const string ProjectAccess = "ProjectAccess"; //Need to allow finance to see projects in addition to the workers, but not enter expenses.
         /// <summary>
         /// When using this Authentication, the projectId must be in the parameters.
         /// </summary>
         public const string PrincipalInvestigator = "PrincipalInvestigator";
         /// <summary>
         /// When using this Authentication, the projectId must be in the parameters.
         /// </summary>
-        public const string PrincipalInvestigatorOnly = "PrincipalInvestigatorOnly ";
+        public const string PrincipalInvestigatorOnly = "PrincipalInvestigatorOnly ";
```
srkirkland, line 22: I know it was here before, but this has a trailing space and we should get rid of it, i think it'll only cause confusion.

Harvest.Web/Controllers/Api/InvoiceController.cs
```diff
@@ -59,7 +59,7 @@ public async Task<ActionResult> Get(int projectId, int invoiceId)
         public async Task<ActionResult> List(int projectId, int? maxRows)
         {
             var user = await _userService.GetCurrentUser();
-            var hasAccess = await _userService.HasAccess(AccessCodes.FieldManagerAccess);
+            var hasAccess = await _userService.HasAccess(AccessCodes.FieldManagerAccess) || await _userService.HasAccess(AccessCodes.FinanceAccess);
```
srkirkland, line 62: 2 things:
1. This doesn't check for team roles, right?
2. Would be nice to extend or overload the HasAccess to check for an array of roles instead of using `||` because this will cause the user's roles to be queried from the Db multiple times, which isn't necessary

jSylvestre: I was thinking about the array of roles...
FYI, it calls this which checks the team:
https://github.com/ucdavis/Harvest/blob/main/Harvest.Core/Services/UserService.cs#L93

srkirkland: @jSylvestre That doesn't check that team as far as I see.  It does return the team name/slug but doesn't check it anywhere.

PR #855, opened 2023-04-19
Jcs/api team route

jSylvestre:


Harvest.Web/Controllers/Api/TicketController.cs
```diff
@@ -152,8 +154,8 @@ public async Task<ActionResult> Create(int projectId, [FromBody] Ticket ticket)
             return Ok(project);
         }
 
-        [HttpGet("/api/[controller]/[action]/{projectId}/{ticketId}")]
         [Authorize(Policy = AccessCodes.PrincipalInvestigator)]
+        [HttpGet("/api/{team}/[controller]/[action]/{projectId}/{ticketId}")]
```
jSylvestre, line 158: Should we be using Route() instead of httpget with params?

PR #856, opened 2023-04-24
When has team route, remove leading slash

jSylvestre:


jSylvestre:
@srkirkland when it isn't using the team route, I left the leading slash in front of the api route.

PR #867, opened 2023-06-02
Upgrade CRA and use MvcReact

### LINKED ISSUE #862: Upgrade to CRA 5.0.1

sprucely:
Closes #862 

The [Node Version] readme badge won't work until it's merged to main.


Harvest.Core/Services/AggieEnterpriseService.cs
```diff
@@ -179,13 +179,13 @@ private async Task GetPpmAccountManager(AccountValidationModel rtValue)
...
         /// <returns></returns>
-        public async Task<FinancialOfficerDetails> GetFinancialOfficer(string financialSegmentString)
+        public Task<FinancialOfficerDetails> GetFinancialOfficer(string financialSegmentString)
         {
             //TODO: If AE can supply the query....
             var rtValue = new FinancialOfficerDetails();
             rtValue.FinancialOfficerId = null;
 
-            return rtValue;
+            return Task.FromResult(rtValue);
```
srkirkland, line 188: I'm not seeing an obvious reason for this, and no code changed seems to be calling this.  Why can't we do this service the regular way?

srkirkland (commented):
Build is failing, but the code changes look good

Harvest.Core/Services/AggieEnterpriseService.cs
jSylvestre: We probably will not implement this for harvest.
It can be deleted if you want.

sprucely: Just getting rid of a compiler complaint about marking a function as async without doing any async. I would have changed the return type from Task<FinancialOfficerDetails> to just FinancialOfficerDetails, but it turned out to be declared in an interface.

jSylvestre (changes requested):
Build failing?

Issue #875, opened 2023-06-23
Overflow issue with FireFox


jSylvestre:
https://harvest-test.azurewebsites.net/caes/Project/Details/70

jSylvestre:
![image](https://github.com/ucdavis/Harvest/assets/469010/aa97845f-99b7-42e3-ac8d-58f99d905cbc)

PR #876, opened 2023-06-23
Another FF wrap

jSylvestre:


jSylvestre:
fixes #875

srkirkland (approved):
don't like that class name but otherwise :rocket:

PR #885, opened 2023-10-18
Maybe a fix for FF not allowing a decimal at the end

jSylvestre:


jSylvestre:
Other pr was a much better fix. Closing.

PR #886, opened 2023-10-19
Rh/firefox decimals

### LINKED ISSUE #884: Other expenses
Currently, Other expenses are limited to whole numbers not decimals.
To get around this, Brian has created other Rates that are $0.10 
Investigate if it is possible to change this to be able to enter decimal amount in the quote and expense.

laholstege:
closes #884 

the issue was that the state of the input could not be changed to something that did not pass parseFloat(). now the state of the input can be erroneous and our onChange() and errors still work the same, the user is just actually allowed to type whatever they want and our parent state will only change to something that passes parseFloat(). 

in firefox this change lets the user type non-numbers into the input, though they don't pass validation. chrome is nice and limits it to numbers. i do not have edge installed lol 

not sure if there's anywhere else that needs a stateful input? @jSylvestre

### DIFF
--- a/Harvest.Core/Data/DbInitializer.cs
+++ b/Harvest.Core/Data/DbInitializer.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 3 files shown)


jSylvestre:
Edge is Chromium so shouldn't be an issue.
I'll test tomorrow.
There is also an acreage input that has the same issue.

jSylvestre (approved):
Nice.
I created a PR to add this to the acres which has the same issue

srkirkland (approved):
Interesting, nice fix

PR #892, opened 2023-11-08
Rh/arcgis map

laholstege:
finally in a readable/workable state! still some bugs and might not be the optimal way of doing things. you can see the map here on the quote page where you're confirming field locations. i've been looking at this: https://localhost:5001/caes/quote/create/104 

there's a map using their built in edit widget: 
image

and one that adds features with react state. we could probably combine this with the edit widget or make our own custom interface. currently i just have a button that adds a predetermined polygon and one that deletes the most recent graphic
image

also was messing with using their sketch widget and their watch() util but that isn't functional. 

pretty interesting, but we should talk about the data storage and what we want to host on ArcGIS (and how). currently we store geometry as a Location column in the db. 

currently, here is what i have hosted for this: 

[Harvest Map](https://ucdavis.maps.arcgis.com/home/item.html?id=8e276b0a2b6b49b88a1ea459d91f5fd3) which contains our basemap of satellite imagery. i don't think we have to specifically add the osm baselayer, but in the toggle we also use an OSM basemap.

[Harvest Fields Feature Layer](https://ucdavis.maps.arcgis.com/home/item.html?id=66289d4facfb4932a5b1d91db8792c4f) where the actual field shapes/graphics are stored.

[Harvest Insight](https://insights.arcgis.com/#/aa56ffae92c44b14a0b4085a32505894) haven't looked into Insights very much or how we would be able to interface it, but pretty cool. i found a map on UCD's ArcGIS Online directory that has Yolo County crop maps and i cross referenced where our field locations are in the Harvest Fields Feature Layer ☝️ and it shows what crop data they have within those locations. that's the workbook but you can also export [a report](https://insights.arcgis.com/#/view/88dea1c9a44a4af7ae02ebfc1d95011a). though you have to manually refresh the data 
image

i can't link to it but i also have an API key generated that we could use to access/query if we need that. seems like we can use the REST API to do [spatial analysis](https://developers.arcgis.com/rest/analysis/api-reference/getting-started.htm) 

also, this is all hosted on my personal account and obviously we need a better system than that. 

my opinion is that this is pretty cool and has a lot more potential than using Leaflet. my biggest questions are around how we want to store/interface with the GIS data, and what access users would have to those layers. also not sure about the licensing/use for GIS services

### DIFF
--- a/Harvest.Web/ClientApp/package-lock.json
+++ b/Harvest.Web/ClientApp/package-lock.json
@@ -8,6 +8,7 @@
       "name": "harvest.web",
       "version": "0.1.0",
       "dependencies": {
+        "@arcgis/core": "^4.28.1",
         "@azure/storage-blob": "^12.14.0",
         "@david.kucsai/react-pdf-table": "^0.4.1",
         "@fortawesome/fontawesome-svg-core": "^1.2.36",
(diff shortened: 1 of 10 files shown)


srkirkland:
Ok -- so my thoughts:
* Looks awesome, I like this approach better than mapbox/leaflet
* Would prefer we keep geoJSON of our fields in the db and dynamically generate the layer, especially at our scale
* I actually like their editor for creating features.  We'd have to handle the onCreate callback and add to our dynamic GeoJSON.
* Not sure the best way to edit features.  Maybe we can use the built in edit, but wasn't clear from this example how to be precise about it.
* I think we if avoid the feature layer, hosting will be simple and we might not even care much/at all about permissions, which would be nice

laholstege:
going to add a geoJSON layer next and get the EditWidget to work with it

PR #896, opened 2023-12-07
Copywrite date

jSylvestre:


Harvest.Web/Views/Error/Index.cshtml
```diff
@@ -130,6 +130,11 @@
             }
         }; 
     </script>
+
+    <script type="text/javascript">
+        //Set copywrite year
+        document.getElementById("copywrite-year").innerHTML = new Date().getFullYear();
```
srkirkland, line 136: @jSylvestre wrong spelling of copyright

srkirkland (changes requested):

Harvest.Web/Views/Error/Index.cshtml
jSylvestre: Yeah, but that says to write the year :)

PR #899, opened 2024-06-26
Fix Typeahead

### LINKED ISSUE #895: Typing in a partial AE COA on the change accounts breaks it. COA button works as expected
Just get a while screen instead of a dropdown

jSylvestre:
Close #895

### DIFF
--- a/Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
+++ b/Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
@@ -185,16 +185,21 @@ export const AccountsInput = (props: Props) => {
         filterBy={() => true} // don't filter on top of our search
         renderMenuItemChildren={(option, propsData, index) => (
           <div>
-            <div>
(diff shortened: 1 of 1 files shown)


Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
```diff
@@ -185,16 +185,21 @@ export const AccountsInput = (props: Props) => {
...
+            {option?.name && (
+              <div>
+                <div>
+                  <Highlighter key="name" search={propsData.text || ""}>
+                    {option.name}
+                  </Highlighter>
+                </div>
+                <div>
+                  <Highlighter key="number" search={propsData.text || ""}>
+                    {option.number}
```
laholstege, line 198: does it break if number is undef too?

jSylvestre: I think if there is text, there will also be a number.
I tested it locally and it now works.

PR #900, opened 2024-06-28
Sigh, that would let them save an empty coa

jSylvestre:


Harvest.Web/ClientApp/src/Requests/AccountsInput.tsx
```diff
@@ -90,6 +90,12 @@ export const AccountsInput = (props: Props) => {
         setError(undefined);
       }
 
+      if (chosenAccount.number === undefined || chosenAccount.number === "") {
```
srkirkland, line 93: replace with `if (!chosenAccount.number) { ... }`

Issue #904, opened 2024-11-27
INC2097603 - show quote amount and billed amount on projects table


jSylvestre:
It would be useful, mostly for completed projects for the PI to see these amounts directly on the table.
Maybe display under the progress bar as quote/amount or amount of quote...
Don't think we want to add another column, but might need to.

jSylvestre:
Need to consider if there is no quote...

PR #905, opened 2024-11-27
Add columns for quote and amount charged

jSylvestre:
close issue #904


sprucely (approved):
10 columns seems like a lot. But I can't think of any simple alternative.

srkirkland (approved):

PR #906, opened 2025-04-07
Jcs/rate validation

jSylvestre:


srkirkland (approved):

Harvest.Core/Services/SlothService.cs
```diff
@@ -599,6 +599,11 @@ private async Task<Result<bool>> ProcessCredits(TransactionViewModel model, deci
                         //Maybe this should prevent it?
                         Log.Warning("Unable to validate credit account {creditAccount}: {creditMessage}", expenseGroup.Key.Account, credit.Message);
                     }
+                    if (!credit.GlSegments.Account.StartsWith("775"))
```
srkirkland, line 602: don't love this but might not be a nicer way.  could put it in options as like 'validPrefix' kinda of thing.  i'll leave it up to you

jSylvestre: Valid rate natural account prefix would be good. Then if they need it to be more than the first 3 chars I can just change that setting.

PR #929, opened 2025-07-22
Prevent zero amount invoices

jSylvestre:


srkirkland (commented):
i think you can get rid of `wasInvoiceCreated` and several if statements if you move the logic inside the `if unbilledExpenses.Any()` and make it a little clearer to follow

PR #953, opened 2025-09-04
Jcs/api middleware v1

jSylvestre:


srkirkland (approved):
this looks pretty good.  One thought is do we want to make team as part of the url, like `/api/{team}/projects` instead of reading the team from the permission?  would be easy as long as you sent that as part of baseURL

jSylvestre:
> this looks pretty good. One thought is do we want to make team as part of the url, like `/api/{team}/projects` instead of reading the team from the permission? would be easy as long as you sent that as part of baseURL

I can, but figured that is all contained in the permission so not really needed.

PR #983, opened 2025-09-22
Adds QR code logos and improves generation

jSylvestre:
Improves QR code generation by adding support for logos.

Introduces new utility functions for creating logos specific to projects and rates, enhancing visual identification.

Updates the QR code generation hook to include options for customizing the logo, such as size, background, and border radius.


jSylvestre:
Not really needed and it might make it so it doesn't decode.

Issue #976, opened 2025-09-16
Change Request Years


jSylvestre:
Need to investigate, but this is probably a bug.

If the change request extends the end data, the new quote doesn't look like it picks up the new years?

<img width="1955" height="884" alt="Image" src="https://github.com/user-attachments/assets/5dfad743-800f-4999-996b-a00766d38213" />

PR #984, opened 2025-09-24
Updates quote years based on project dates.

jSylvestre:
Ensures the quote reflects the correct number of years based on the project's start and end dates.  If the project duration has changed since the quote was created, an alert informs the user and updates the quote's year value.

## Summary by CodeRabbit

* **New Features**
  * Quotes now auto-recalculate project duration (minimum 1 year) on load and alert users when the value changes.
  * Change-requested projects show an Original Timeline when original start/end dates differ; a loading message appears while fetching.
  * Approving change requests that modify project dates now records a history entry for auditing.

* **Style**
  * Danger-text within page-wrapped content has improved visibility via updated color.


jSylvestre:
Close #976

PR #995, opened 2025-10-20
Adds basic iPhone login

jSylvestre:
Adds an initial attempt at iPhone login using a simple key check.

Currently, it retrieves a user with a hardcoded Kerberos value. It also limits user authorization to CAES team members.

## Summary by CodeRabbit

* **New Features**
  * Added iPhone app authentication support for Apple App Review login flow, issuing a short-lived token and redirecting to the mobile app for review sessions.
* **Chores**
  * Added a new configuration key to enable the iPhone review flow.
  * Updated development routing to expose the review endpoint during testing.

### DIFF
--- a/Harvest.Web/Controllers/AppleReviewController.cs
+++ b/Harvest.Web/Controllers/AppleReviewController.cs
@@ -60,12 +60,19 @@ namespace Harvest.Web.Controllers
                 return NotFound();
             }
 
+            var baseUrl = $"{Request.Scheme}://{Request.Host}";
+            //Check base url is for test only: https://harvest-test.azurewebsites.net
+            if(!baseUrl.Equals("https://harvest-test.azurewebsites.net", StringComparison.OrdinalIgnoreCase) )
+            {
+                return BadRequest("This endpoint can only be used in test environment");
+            }
+
             permission.Token = Guid.NewGuid();
             permission.TokenExpires = DateTime.UtcNow.AddMinutes(5);
 
             await _dbContext.SaveChangesAsync();
 
-            var baseUrl = $"{Request.Scheme}://{Request.Host}";
+            
             var appLink = $"harvestmobile://applink?code={permission.Token}&baseUrl={Uri.EscapeDataString(baseUrl)}";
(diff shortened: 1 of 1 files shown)


Harvest.Web/Controllers/AppleController.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Threading.Tasks;
+
+namespace Harvest.Web.Controllers
+{
+    public class AppleController : Controller
+    {
+        private readonly AppDbContext _dbContext;
+        private readonly AuthSettings _authSettings;
+
+        public AppleController(AppDbContext dbContext, IOptions<AuthSettings> authSettings)
```
srkirkland: maybe call it `AppleReviewController` or `MobileReviewController` because i think i'll forget what an AppleController is after a while

Harvest.Web/Controllers/AppleController.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            _dbContext = dbContext;
+            _authSettings = authSettings.Value;
+            if (string.IsNullOrEmpty(_authSettings.IPhoneKey) || _authSettings.IPhoneKey == "[External]")
```
srkirkland: better to put inside the login method instead of ctor?  I'm worried it could break prod if we don't remember to add in the config instead of just breaking this one method

Harvest.Web/Controllers/AppleReviewController.cs
```diff
@@ -0,0 +1,67 @@
...
+    public class AppleController : Controller
+    {
+        private readonly AppDbContext _dbContext;
+        private readonly AuthSettings _authSettings;
+
+        public AppleController(AppDbContext dbContext, IOptions<AuthSettings> authSettings)
+        {
+            _dbContext = dbContext;
+            _authSettings = authSettings.Value;
+            if (string.IsNullOrEmpty(_authSettings.IPhoneKey) || _authSettings.IPhoneKey == "[External]")
```
jSylvestre: Done, but controller got renamed, so isn't showing up here

